PR #5, opened 2019-05-31
Optimize image pushes and integration tests

anandswaminathan:
- Currently we build and push images on all PRs
- Prevents images to be pushed on all PRs.

This PR will make images to be pushed only after merging to master. Integration tests will build images, and use the locally built images to run tests. 

We can in future make images to be pushed to dockerhub on successful PR if needed.


integ/main_test.go
```diff
@@ -36,8 +36,8 @@ func (s *IntegSuite) SetUpSuite(c *C) {
 	}
 	// var runDirect = flag.Bool("runDirect", false, "if set, runs the operator from the current source instead of from an image")
 	var runDirect = os.Getenv("RUN_DIRECT") != ""
-	// var image = flag.String("image", "lyft/flinkk8soperator:latest", "image for the operator")
-	var image = os.Getenv("OPERATOR_IMAGE")
+	// var image = flag.String("image", "flinkk8soperator:latest", "image for the operator")
+	var image = os.Getenv("IMAGE")
```
mwylde: I think OPERATOR_IMAGE is clearer, as there are several different images involved.

anandswaminathan: Yes. The documentation was IMage. Will change to Operator_image everywhere

mwylde: Yeah, I forgot to update the doc when I changed it a while back.

PR #7, opened 2019-06-03
Add node selectors to job and task config

annymsMthd:

### DIFF
--- a/pkg/apis/app/v1alpha1/types.go
+++ b/pkg/apis/app/v1alpha1/types.go
@@ -101,6 +101,7 @@ type JobManagerConfig struct {
 	Environment           EnvironmentConfig           `json:"envConfig"`
 	Replicas              *int32                      `json:"replicas,omitempty"`
 	OffHeapMemoryFraction *float64                    `json:"offHeapMemoryFraction,omitempty"`
+	NodeSelector          map[string]string           `json:"nodeSelector"`
 }
 
 type TaskManagerConfig struct {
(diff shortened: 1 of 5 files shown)


annymsMthd:
Looks like the `make dockerhub_push` should be removed from PRs.

anandswaminathan:
@annymsMthd Just pushed a PR to remove docker_hub push from PRs.

pkg/apis/app/v1alpha1/types.go
```diff
@@ -101,13 +101,15 @@ type JobManagerConfig struct {
 	Environment           EnvironmentConfig           `json:"envConfig"`
 	Replicas              *int32                      `json:"replicas,omitempty"`
 	OffHeapMemoryFraction *float64                    `json:"offHeapMemoryFraction,omitempty"`
+	NodeSelector          map[string]string           `json:"nodeSelector"`
```
anandswaminathan: Can you add "omitempty"  like above to prevent serialization when empty ?

annymsMthd:
@anandswaminathan should be gtg now

mwylde:
Hey Joshua — can you give a bit more context on why this is needed?

annymsMthd:
Sure thing. In our kubernetes cluster we have different labels for nodes. Since we self host we have dedicated nodes for certain workloads. e.g. ingress, worker, storage, etc.

annymsMthd:
@mwylde @anandswaminathan any reason not to move forward with this? This is a pretty common thing in k8s.

PR #9, opened 2019-06-03
Increase flink api timeout to avoid submitting duplicate jobs

annymsMthd:

### DIFF
--- a/pkg/controller/flink/client/api.go
+++ b/pkg/controller/flink/client/api.go
@@ -30,7 +30,7 @@ const httpGet = "GET"
 const httpPost = "POST"
 const httpPatch = "PATCH"
 const retryCount = 3
-const timeOut = 5 * time.Second
+const timeOut = 1 * time.Minute
 
 type FlinkAPIInterface interface {
 	CancelJobWithSavepoint(ctx context.Context, url string, jobID string) (string, error)


annymsMthd:
Not sure if a minute is a good number here but I was getting this message:
```
Events:
  Type     Reason  Age    From              Message
  ----     ------  ----   ----              -------
  Normal   Create  10m    flinkk8soperator  Flink cluster created
  Warning  Create  8m56s  flinkk8soperator  Failed to submit job to cluster: Submit job API request failed: Post http://myapp-27e1e0ef.default:8081/jars/myjar.jar/run: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```
And eventually there was 3 of the same job running.

Running this change has the correct behavior with no timeout and only one job running

anandswaminathan:
@annymsMthd 

Hey, the reason we brought down api timeouts is because this is blocking call and holds up a go routine. The side effects are bad when pods are up and job manager is not able to respond to API requests. Meaning all the routines will be blocked until job manager can respond to requests. 

What version of Flink did you try out? Wondering why Jobmanager is taking greater than 5 seconds to accept job submission requests.

anandswaminathan:
Also this problem will happen again if there is timeout after a min. We need to remove retries from the resty client for Post requests I believe. That way the controller would check for jobs running before submitting in the subsequent callback.

annymsMthd:
Ah I see. Yeah that sounds good.

annymsMthd:
We are using the same Flink version as the example in this project 1.8.0. It looks like it takes around 11.8 seconds to submit the job.

anandswaminathan:
@annymsMthd @mwylde Planning to create a Github issue for this one as I believe based on the underlying application we might encounter this more, and address it this way.

Only Get, patch requests will have a Retry (of 3) and time out of 5 seconds.
For post - There will be no be retries at the Resty client layer, and have a high timeout of 1 minute. This might not be needed if Flink has some form of idempotency token for job submission as well. For now, we will still need to address it.

anandswaminathan:
@annymsMthd I just realized that Job submission takes a long time when CPU and memory allocated to Job manager is low. Can you try to run the application with higher CPU/memory and see if the job submission still takes 11 seconds? Ideally it should finish within a second.

annymsMthd:
testing this out now

annymsMthd:
@anandswaminathan Updating the cpu resources for the job manager to 1 second from 500 milliseconds and now it completes in 6.411695194s

YuvalItzchakov:
I encountered this as well:

```
RESTY 2019/06/06 12:08:53 ERROR Post http://foo-e8cfa581.flink-operator:8081/jars/foo.jar/run: net/http: request canceled (Client.Timeout exceeded while awaiting headers), Attempt 1
RESTY 2019/06/06 12:08:53 ERROR Post http://foo-e8cfa581.flink-operator:8081/jars/foo.jar/run: net/http: request canceled (Client.Timeout exceeded while awaiting headers), Attempt 2
RESTY 2019/06/06 12:08:53 ERROR Post http://foo-e8cfa581.flink-operator:8081/jars/foo.jar/run: net/http: request canceled (Client.Timeout exceeded while awaiting headers), Attempt 3
```

And ended up with 3 jobs running. I think it would be useful to make this configurable to the users as you never know under what limitations (resources and time) their cluster will be under.

anandswaminathan:
Merged https://github.com/lyft/flinkk8soperator/pull/28

PR #10, opened 2019-06-03
Allow annotations for the ingress at the app level

annymsMthd:

### DIFF
--- a/pkg/apis/app/v1alpha1/types.go
+++ b/pkg/apis/app/v1alpha1/types.go
@@ -35,6 +35,7 @@ type FlinkApplicationSpec struct {
 	FlinkVersion      string                       `json:"flinkVersion"`
 	TaskManagerConfig TaskManagerConfig            `json:"taskManagerConfig,omitempty"`
 	JobManagerConfig  JobManagerConfig             `json:"jobManagerConfig,omitempty"`
+	IngressConfig     IngressConfig                `json:"ingressConfig,omitempty"`
 	JarName           string                       `json:"jarName"`
 	Parallelism       int32                        `json:"parallelism"`
 	EntryClass        string                       `json:"entryClass,omitempty"`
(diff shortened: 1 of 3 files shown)


annymsMthd:
Our k8s clusters have multiple ingress controllers so it would be helpful to allow annotations for ingresses:
```
ingressConfig:
    annotations:
      kubernetes.io/ingress.class: "private"
```

pkg/apis/app/v1alpha1/types.go
```diff
@@ -35,6 +35,7 @@ type FlinkApplicationSpec struct {
 	FlinkVersion      string                       `json:"flinkVersion"`
 	TaskManagerConfig TaskManagerConfig            `json:"taskManagerConfig,omitempty"`
 	JobManagerConfig  JobManagerConfig             `json:"jobManagerConfig,omitempty"`
+	IngressConfig     IngressConfig                `json:"ingressConfig,omitempty"`
```
anandswaminathan: @annymsMthd 

Right now for an application when an Ingress is created, it is never updated (unlike Service and Deployment). Let's say you add an annotation for Ingress config, and you update it, the update will never be reflected in the Ingress.

annymsMthd: Okay. I can add that if you guys like this addition to the api.

anandswaminathan: Will have to think it through. We currently do not have this requirement.

We pass on the annotation from the flink application to the deployments. Would doing something similar for ingress and services address your use case?

annymsMthd: That would work.

anandswaminathan: Sorry for the delay. Forgot about this thread.

Submitted a PR: https://github.com/lyft/flinkk8soperator/pull/74

PR #11, opened 2019-06-03
Log operator config on startup

glaksh100:


cmd/flinkk8soperator/cmd/root.go
```diff
@@ -140,6 +140,19 @@ func executeRootCmd(controllerCfg *controller_config.Config) error {
 	}
 }
 
+func logConfigMap(ctx context.Context, controllerCfg *controller_config.Config) {
+	logger.Infof(ctx, "Staleness Duration %v", controllerCfg.StatemachineStalenessDuration)
```
mwylde: Is it possible to use reflection to do this dynamically? That might lessen the work of keeping this in sync with the config as it evolves.

anandswaminathan: +1

I believe you can simply do `logger.Infof(ctx, "%+v", controllerCfg)`

glaksh100: Thanks y'all. Updated to use `logger.Infof(ctx, "%+v\n", controllerCfg)`

glaksh100:
🚀

PR #12, opened 2019-06-03
Added finalizer permissions

yuchaoran2011:
1. I tried out the project and found out an issue with starting a Flink app:
```
time="2019-06-03T18:10:48Z" level=error msg="Jobmanager deployment creation failed deployments.apps \"wordcount-operator-example-f5b2ddb2-jm\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: no RBAC policy matched, <nil>" app_name=wordcount-operator-example ns=flink-operator phase=
```
The reason is that the finalizer permissions were not properly given to the ClusterRole. This PR fixed it. (Note: I'm on OpenShift. Not sure if it's OpenShift-specific but this change shouldn't break non-OpenShift k8s).

2. Added some notes about resource requirements when deploying the operator. I wasn't able to deploy initially because of lack of resources. I think it's better to let the user know that the default requirements might not be what he wants.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -37,7 +37,7 @@ rules:
     - create
     - update
     - delete
-#Allow Event recording access
+# Allow Event recording access
  - apiGroups:
     - ""
    resources:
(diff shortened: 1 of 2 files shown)


anandswaminathan:
Thanks for fixing. Merging this to master.

PR #17, opened 2019-06-04
Minor fixes and improvements

yuchaoran2011:
1. Updated Go to the most recent version (i.e. 1.12.5). 1.10.4 is way too old.
2. Updated the URL for S2I. The previous URL was pointing to a location that redirected to the current URL.
3. It's always good to specify the namespace for a namespaced resource. So a namespace `default` is added to the sample FlinkApplication YAML and READMEs are updated accordingly. Previously, without namespace specified, the CR object was also created in `default` namespace by default. But the README was assuming it was in the `flink-operator` namespace.
4. Changed `jobmanager.web.upload.dir` to `web.upload.dir`. The former is now deprecated.

### DIFF
--- a/docs/quick-start-guide.md
+++ b/docs/quick-start-guide.md
@@ -74,7 +74,7 @@ $ kubectl get deployments -n flink-operator
 
 Check the phase and other status attributes in the custom resource
 ```bash
-$ kubectl get flinkapplication.flink.k8s.io -n flink-operator wordcount-operator-example -o yaml
+$ kubectl get flinkapplication.flink.k8s.io -n default wordcount-operator-example -o yaml
 ```
 
 The output should be something like this
@@ -135,7 +135,7 @@ status:
 To check events for the `FlinkApplication` object, run the following command:
 
 ```bash
-$ kubectl describe flinkapplication.flink.k8s.io -n flink-operator wordcount-operator-example
+$ kubectl describe flinkapplication.flink.k8s.io -n default wordcount-operator-example
 ```
 
 This will show the events similarly to the following:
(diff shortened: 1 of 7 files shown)


docs/quick-start-guide.md
```diff
@@ -74,7 +74,7 @@ $ kubectl get deployments -n flink-operator
 
 Check the phase and other status attributes in the custom resource
 ```bash
-$ kubectl get flinkapplication.flink.k8s.io -n flink-operator wordcount-operator-example -o yaml
+$ kubectl get flinkapplication.flink.k8s.io -n default wordcount-operator-example -o yaml
```
anandswaminathan: The application is created in `flink-operator` above

yuchaoran2011: Ok I missed that. I updated the create command to remove the `-n` flag. I think It's better to let the YAML of a namespaced resource (a `FlinkApplication` in this case) specify its namespace. This removes opportunities of mis-creating it in the wrong namespace and avoids potential RBAC issues.

anandswaminathan: Sorry, I am not sure I follow your comment. I am still wondering why the application is moved to `default` namespace from `flink-operator`. We can just add the namespace `flink-operator` in the yaml file right ?

In the current quick start guide, all the operator resources are in `flink-operator`. To maintain consistency, we added the application also in the same namespace just for quick start. Note that, anyone is free to create applications in any namespace, as long as Operator is monitoring that particular namespace.

yuchaoran2011: Oh I just meant that in the sample app YAML, we should have namespace specified. In which namespace doesn't matter. I'll move it back to the `flink-operator` namespace now that you think it's better to be there.

anandswaminathan:
Thanks. Will merge this tonight.

PR #23, opened 2019-06-11
Handle invalid phases in the Flink state machine

anandswaminathan:
The operator currently panics if it receives an application callback with an invalid phase.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -98,6 +98,11 @@ func (s *FlinkStateMachine) shouldRollback(ctx context.Context, application *v1a
 
 func (s *FlinkStateMachine) Handle(ctx context.Context, application *v1alpha1.FlinkApplication) error {
 	currentPhase := application.Status.Phase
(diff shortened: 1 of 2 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -98,6 +98,11 @@ func (s *FlinkStateMachine) shouldRollback(ctx context.Context, application *v1a
 
 func (s *FlinkStateMachine) Handle(ctx context.Context, application *v1alpha1.FlinkApplication) error {
 	currentPhase := application.Status.Phase
+	if _, ok := s.metrics.stateMachineHandlePhaseMap[currentPhase]; !ok {
+		errMsg := fmt.Sprintf("Invalid state %s for the application", currentPhase)
+		logger.Errorf(ctx, errMsg)
```
mwylde: Do we need to explicitly log this here? Looks like the error returned from Handle is already logged in Reconcile.

anandswaminathan: oh sure. I can remove it.

mwylde:
👍

PR #26, opened 2019-06-13
Made sample app run on OpenShift

yuchaoran2011:
1. Updated the Dockerfile and entrypoint script to address permission issues in order for the sample apps to run on Openshift. The changes shouldn't affect other K8s flavors.
2. Miscellaneous updates to package versions (Go lang, gosu, Maven and Flink's Scala).

Tested on both OpenShift and GKE.


anandswaminathan:
👍 

Will build the new docker image for Wordcount  example and associate latest tag to it.

PR #27, opened 2019-06-13
[STRMCMP-509] Make old cluster clean-up idempotent

mwylde:
Currently, if we successfully delete just one of the two (TM and JM) deployments during old-cluster clean-up, on the next run we will not be able to successfully query the deployment due to the check in GetCurrentAndOldDeploymentsForApp that we have 1 JM and 1 TM for each. This means that we will never clean up the old deployment. This may also cause us to not clean up the old versioned service.

This PR refactors the cleanup process to query for all old resources (deployments and services) and delete them.

Note that this is a (slightly) incompatible change. As we now rely on labels to find old services that need to be cleaned up, this operator will not be able to clean up services created by a previous version—they will need to be deleted manually after the upgrade.

### DIFF
--- a/pkg/controller/flink/flink.go
+++ b/pkg/controller/flink/flink.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"github.com/lyft/flinkk8soperator/pkg/controller/common"
 
 	"github.com/lyft/flinkk8soperator/pkg/controller/config"
@@ -45,9 +47,6 @@ type ControllerInterface interface {
 	// Creates a Flink cluster with necessary Job Manager, Task Managers and services for UI
 	CreateCluster(ctx context.Context, application *v1alpha1.FlinkApplication) error
 
-	// Deletes a Flink cluster based on the hash
-	DeleteCluster(ctx context.Context, application *v1alpha1.FlinkApplication, hash string) error
-
 	// Cancels the running/active jobs in the Cluster for the Application after savepoint is created
 	CancelWithSavepoint(ctx context.Context, application *v1alpha1.FlinkApplication, hash string) (string, error)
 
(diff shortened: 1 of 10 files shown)


anandswaminathan:
@mwylde I am wondering what is the reasoning behind having the check in `GetCurrentAndOldDeploymentsForApp` - I guess "len(ds) != 2" in "listToFlinkDeployment" is the issue right? Removing that check will fix everything right? 

If needed we can just have that check for current deployments for added protection. Labels are tricky. Anyone can add any labels in Kubernetes. Deleting by just labels could be dangerous. The chances are low but why take it.

mwylde:
We want to ensure that the current deployment returned from that function always has a TM and JM—otherwise something has gone very wrong. That always doesn't solve the service case. I think separating our these two functions (finding the current deployments, and finding old resources that need to be cleaned up) is cleaner than how I was doing it before.

mwylde:
As far as using labels, the issue is that we _don't_ know the names during the deletion process, as for that we need to know previous hashes.

pkg/controller/flink/flink.go
```diff
@@ -348,52 +317,84 @@ func listToFlinkDeployment(ds []v1.Deployment, hash string) *common.FlinkDeploym
...
-	if cur != nil && application.Status.FailedDeployHash == "" &&
-		(!f.deploymentMatches(ctx, cur.Jobmanager, application) || !f.deploymentMatches(ctx, cur.Taskmanager, application)) {
-		// we had a hash collision (i.e., the previous application has the same hash as the new one)
-		// this is *very* unlikely to occur (1/2^32)
-		return nil, nil, errors.New("found hash collision for deployment, you must do a clean deploy")
+	oldObjects := make([]metav1.Object, 0)
+
+	for _, d := range deployments.Items {
+		if d.Labels[FlinkAppHash] != curHash {
+			oldObjects = append(oldObjects, d.DeepCopy())
```
anandswaminathan: Before appending to oldObjects,

Since you have the and the application hash, can you do a name check for jm/tm ?

pkg/controller/flink/flink.go
```diff
@@ -348,52 +317,84 @@ func listToFlinkDeployment(ds []v1.Deployment, hash string) *common.FlinkDeploym
...
-	cur := listToFlinkDeployment(byHash[curHash], curHash)
-	if cur != nil && application.Status.FailedDeployHash == "" &&
-		(!f.deploymentMatches(ctx, cur.Jobmanager, application) || !f.deploymentMatches(ctx, cur.Taskmanager, application)) {
-		// we had a hash collision (i.e., the previous application has the same hash as the new one)
-		// this is *very* unlikely to occur (1/2^32)
-		return nil, nil, errors.New("found hash collision for deployment, you must do a clean deploy")
+	oldObjects := make([]metav1.Object, 0)
+
+	for _, d := range deployments.Items {
+		if d.Labels[FlinkAppHash] != curHash {
```
anandswaminathan: Can you also add extra check label map with has key `FlinkAppHash`, and also check if value is not empty string.

anandswaminathan:
@mwylde Makes sense. Just a comment to add extra protection. I don't want us to get rid of all deployments if they have a label `flink-app`

pkg/controller/flink/flink.go
mwylde: Yeah, that's reasonable. We might also want to use namespaced labels (`flink.k8s.io/flink-app`, etc.) to reduce the chance of collisions.

mwylde:
Updated with the additional name and tag validation

anandswaminathan:
👍

mwylde:
🚀

Issue #13, opened 2019-06-03
Flink cluster got into a bad state


yuchaoran2011:
I'm not able to delete the FlinkApplication CR object:
```
$ $ oc delete FlinkApplication wordcount-operator-example
flinkapplication.flink.k8s.io "wordcount-operator-example" deleted
$ oc get FlinkApplication
NAME                         AGE
wordcount-operator-example   28m
```
As shown above, the `delete` command did nothing.

The pods associated with the FlinkApp are not cleaned up either:
<img width="1261" alt="Screen Shot 2019-06-03 at 16 28 02" src="https://user-images.githubusercontent.com/1168769/58832279-9c3d9a80-861c-11e9-996a-e94c74df8b87.png">

YuvalItzchakov:
I ran into the same issue. The operator is unable to delete the underlying deployments if they're in a bad state. A workaround is to run:

```bash
kubectl patch flinkapplications/{APPLICATION_NAME} -p '{"metadata":{"finalizers":[]}}' --type=merge
```

Which causes all finalizers to reset and quicks off termination.

anandswaminathan:
@YuvalItzchakov @yuchaoran2011 Thanks for reporting. This happens because the example we have in the quick start guide completes. If you have a streaming application, that does not move to get "FINISHED", this would not happen. We will get a fix for this.

@mwylde The resource is stuck in Deletion as we cannot savepoint as there is no flink job running in the cluster. Hence the finalizer is not getting cleared. This happens in the following case:

1) If underlying Flink Job finishes.
2) If the underlying Flink job was cancelled from outside the Operator.
3) If after issuing CancelWithSavepoint, we failed to store the `triggerId` in the custom resource due to failures. (3 is extremely rare but is possible)

What do you think about clearing the finalizer if  jobFinished is true here: https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L529

yuchaoran2011:
@anandswaminathan Thanks for the reply. Actually the example app is a streaming app (because it's adapted from this [example](https://github.com/apache/flink/blob/master/flink-examples/flink-examples-streaming/src/main/java/org/apache/flink/streaming/examples/wordcount/WordCount.java)) but it still completes.

I think the reason that it completes is because the input data is from a bounded data source. In other words, all batch apps and streaming apps with a bounded data source will complete. Only streaming apps with an unbounded data source continues indefinitely.

PR #30, opened 2019-06-18
[STRMCMP-544] Allow FlinkApplications to be deleted when job submission fails

mwylde:
Currently, if a user creates a new FlinkApplication for which we can't successfully submit the job (e.g., the jar name is wrong, the job crashes on startup, the specified savepoint doesn't exist, etc.) the user will not be able to then delete the FlinkApplication without changing the DeleteMode to None. This is because we do not set the DeployHash until after job submission, which prevents us from querying the versioned cluster. This PR addresses that by allowing immediate deletion of a cluster when the DeployHash is not set.

It also addresses the issue in #13 by allowing deletion when there has been a successful deploy at some point, but at delete time no jobs are running.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -517,14 +517,21 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
 	// If https://github.com/kubernetes/kubernetes/issues/56567 is fixed users will be able to use
 	// kubectl delete --force, but for now they will need to update the DeleteMode.
 
-	if app.Spec.DeleteMode == v1alpha1.DeleteModeNone {
-		// just delete the finalizer so the cluster can be torn down
+	// If the delete mode is none or there's no deployhash set (which means we failed to submit the job on the
+	// first deploy) just delete the finalizer so the cluster can be torn down
+	if app.Spec.DeleteMode == v1alpha1.DeleteModeNone || app.Status.DeployHash == "" {
 		return s.clearFinalizers(ctx, app)
 	}
 
 	jobs, err := s.flinkController.GetJobsForApplication(ctx, app, app.Status.DeployHash)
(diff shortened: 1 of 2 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -527,6 +528,12 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
 		return err
 	}
 
+	if len(jobs) == 0 {
```
anandswaminathan: Shouldn't this be `if finished ` ? -  based on the comment below ?

mwylde: No, we already handle that case in jobFinished. This handles the case that the job finished or was canceled and the jobmanager restarted, in which case the job will no longer be returned from the API.

anandswaminathan: @mwylde Check this issue - https://github.com/lyft/flinkk8soperator/issues/13

This is scenario - DeleteMode = DeleteModeSavepoint, the underlying job is finished or cancelled. The deletion is stuck.

I hit this case when migrating to controller runtime, where we failed to store the triggerId in etcd. In the next loop, the job was already cancelled.

mwylde: Sorry, you're totally right. That case is now handled correctly (with a test).

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -528,6 +529,12 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
 	}
 
 	finished := jobFinished(jobs, app.Status.JobStatus.JobID)
+
+	if len(jobs) == 0 || finished {
```
anandswaminathan: +1

Can you remove code beneath - line 540, and 549 ? They are redundant.

PR #31, opened 2019-06-18
Added support for job cluster

yuchaoran2011:
Added support for job cluster mode.

As discussed with @anandswaminathan last Friday, it's important to make the Flink app configurable (e.g. by passing application arguments and configuring the resource requirements such as level of parallelism) without having to rebuild the application image. The job cluster mode has the same benefits as the session cluster mode, which is what's currently supported in the operator, while also being simpler. Because in the job cluster mode, no job submission is required (hence no REST API needs to be invoked for that purpose), the user also can place his application jar anywhere on the Flink classpath (as opposed to a fixed location configured by the `web.upload.dir` property.

This PR adds an option in the CR YAML file so that the user can choose to run the job in either session cluster mode or the job cluster mode. Session cluster mode is the default if no cluster mode is specified.


pkg/controller/flink/job_manager_controller.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/lyft/flytestdlib/promutils/labeled"
 	v1 "k8s.io/api/apps/v1"
 	coreV1 "k8s.io/api/core/v1"
-	k8_err "k8s.io/apimachinery/pkg/api/errors"
+	k8err "k8s.io/apimachinery/pkg/api/errors"
```
yuchaoran2011: This is a minor style thing suggested by my editor. So I went ahead and fixed it. Snake case variable names are discouraged in Go.

PR #32, opened 2019-06-18
Improve staleness logic by improving error handling and retrying a subset of errors

glaksh100:
This PR removes dependency on staleness duration. The application is now rolled back based on the error observed (and retried if applicable). It also makes error handling a bit more structured so that we can more easily specify retry-able errors.

**Changelog**
- Adds a error_handler class that wraps around the error interface and adds information like error codes and method info to the error message
- Defines a map of retryable error keys (method + error code)
- Unifies all error logging in api.go to use the above error_handler
- Adds a `LastSeenError` and a `RetryCount` to the FlinkApplicationStatus
- Changes the logic within `shouldRollback()` in flink_state_machine.go to check if the LastSeenError can be retried and is within the maxRetries limit
- Removes dependency on staleness duration

**Note**: go generate on Config.go seems to be failing with:
```
which pflags || (go get github.com/lyft/flytestdlib/cli/pflags)
# github.com/lyft/flytestdlib/promutils
../flytestdlib/promutils/workqueue.go:29:49: cannot use composite literal (type prometheusMetricsProvider) as type workqueue.MetricsProvider in argument to workqueue.SetProvider:
	prometheusMetricsProvider does not implement workqueue.MetricsProvider (missing NewDeprecatedAddsMetric method)
make: *** [gen-config] Error 2
```


glaksh100:
Many thanks to @anandswaminathan for helping debug and fix the pflags issue. For anyone who may run into this in their local dev:
- make gen-config downloads `flytestdlib` to ~/src/go/blah...`  (does a `go get`) 
- All references to flytestdlib are henceforth to your local flystdlib version and NOT to the one under `flinkk8soperator/vendor/github.com/lyft/flytestdlib`. 
-  **TL;DR** run dep ensure on your local version of flytstdlib and the problem goes away.

glaksh100:
@mwylde @anandswaminathan After various fixes,  I think this PR is now in a state to be reviewed. PTAL when you have a chance.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -77,23 +78,30 @@ func (s *FlinkStateMachine) updateApplicationPhase(ctx context.Context, applicat
...
-				elapsedTime, application.Status.Phase))
-
-			return true
-		}
+	if application.Status.LastSeenError == "" {
+		return false
 	}
-	return false
+	// Check if the error is retryable and we have retries left
+	if client.IsErrorRetryable(application.Status.LastSeenError) && application.Status.RetryCount <= maxRetries {
```
mwylde: Is there anyway to implement exponential backoff for retries? Otherwise we may very quickly exhaust our retries without giving the problem time to resolve itself.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -181,6 +190,7 @@ func (s *FlinkStateMachine) handleClusterStarting(ctx context.Context, applicati
 	// Wait for all to be running
 	ready, err := s.flinkController.IsClusterReady(ctx, application)
 	if err != nil {
+		s.getAndUpdateError(ctx, application, err)
```
mwylde: In ClusterStarting, there's a class of errors with the underlying deployments (or pods) that could prevent the cluster from coming up (for example, a bad image). I think we'll still need some kind of time limit to handle that case, as currently the deployment will just get stuck indefinitely.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -156,6 +164,7 @@ func (s *FlinkStateMachine) handleNewOrUpdating(ctx context.Context, application
 	// Create the Flink cluster
 	err := s.flinkController.CreateCluster(ctx, application)
 	if err != nil {
+		s.getAndUpdateError(ctx, application, err)
```
mwylde: This pattern seems a bit error-prone to me, as it requires us to remember to include this before every return err in order to avoid potentially getting stuck (for example, these are missing from several places in handleApplicationSavepointing).

Can we do for all errors in one place?

anandswaminathan:
@glaksh100 

One thing I have noticed with both Operator SDK and Controller runtime is that on occasions - Callbacks can happen at a much faster loop, and not only at the `Resync` period. Have tried debugging it but looks like several edge cases where this can happen. 

What this can translate to - you might exhaust your maxRetries in very small interval. You will need to track time, and have logic like `minTimeBetweenRetries` or something, and set it to an acceptable levels.

pkg/controller/flink/client/error_handler.go
```diff
@@ -63,3 +69,46 @@ func GetErrorKey(error error) string {
...
+type RetryHandler struct {
+	baseBackOffMillis int32
+	timeToWaitMillis  int32
+}
+
+func NewRetryHandler(baseBackoff int32, timeToWait int32) RetryHandler {
+	return RetryHandler{baseBackoff, timeToWait}
+}
+func (r RetryHandler) IsErrorRetryable(err string) bool {
+	if _, ok := retryableErrors[err]; ok {
```
mwylde: This worries me a bit, because typo'ing the name in this map (or updating the error message but not updating the map) would cause this to silently fail. It might be less error-prone to have the map include all errors along with a boolean indicating whether they're retryable; then we could error from this method if an exception isn't found.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -77,22 +80,46 @@ func (s *FlinkStateMachine) updateApplicationPhase(ctx context.Context, applicat
 }
 
 func (s *FlinkStateMachine) shouldRollback(ctx context.Context, application *v1alpha1.FlinkApplication) bool {
-	if application.Status.DeployHash == "" {
+	if application.Status.DeployHash == "" && application.Status.LastSeenError == "" {
```
mwylde: Adding this condition doesn't make sense to me — the idea behind this check was that if this is the first deploy we can't roll back because there's nothing to roll back _to_, so we might as well just keep trying.

pkg/controller/flink/client/error_handler.go
```diff
@@ -0,0 +1,114 @@
...
+	return false
+}
+
+func (r RetryHandler) WaitOnError(clock clock.Clock, lastUpdatedTime time.Time) (time.Duration, bool) {
+	elapsedTime := clock.Since(lastUpdatedTime)
+	return elapsedTime, elapsedTime <= time.Millisecond*time.Duration(r.timeToWaitMillis)
+
+}
+func (r RetryHandler) GetRetryDelay(retryCount int32) time.Duration {
+	return time.Duration(1<<uint(retryCount)*r.baseBackOffMillis) * time.Millisecond
```
mwylde: We probably want to bound this to some reasonable value (32 seconds)?, as exponential backoffs can get ridiculous pretty quickly 😄. It can also be good to add some random jitter to avoid everybody retrying at the same time (for example due to a kubeapi issue that affects all apps).

pkg/controller/flink/client/error_handler.go
```diff
@@ -0,0 +1,114 @@
...
+}
+func (r RetryHandler) BackOff(retryCount int32) {
+	time.Sleep(r.GetRetryDelay(retryCount))
```
mwylde: @anandswaminathan can you comment on the implications of sleeping the goroutine? 

It might be better to encode this in the FlinkApplication (or maybe you can already determine it based on the information there?). Then you could check whether it's time to retry as part of the normal Handle call.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -22,7 +22,9 @@ import (
 )
 
 const (
-	jobFinalizer = "job.finalizers.flink.k8s.io"
+	jobFinalizer           = "job.finalizers.flink.k8s.io"
+	retryBaseBackoffMillis = 100
+	timeToWaitMillis       = 5 * 60 * 1000
```
mwylde: Should these be configurable? If the timeToWaitMilis is used like the existing staleness duration then it may need to be tuned based on e.g., how fast the cluster can spin up pods.

pkg/controller/flink/client/error_handler.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/clock"
+)
+
+// appError codes
+const (
+	globalFailure      = "FAILED"
+	jsonUnmarshalError = "JSONUNMARSHALERROR"
+	defaultRetries     = 5
```
mwylde: I think this, combined with the 100ms backoff, is too low. I think this gives a retry schedule like

```
0	100  	100
1	200  	300
2	400 	700
3	800 	1500
4	1600	3100
```
(count, backoff ms, cumulative ms)

So max in total we only wait 3 seconds. I recommend something like 20 retries which (combined with a bound on the backoff) could give a failing external resource time to recover.

glaksh100:
@mwylde Thanks for the initial and the second review! I'm going to respond to comments individually and resolve ones that are outdated.

pkg/controller/flink/client/api.go
```diff
@@ -217,18 +217,17 @@ func (c *FlinkJobManagerClient) SubmitJob(ctx context.Context, url string, jarID
...
 		c.metrics.submitJobFailureCounter.Inc(ctx)
-		return nil, errors.Wrap(err, "Submit job API request failed")
+		return nil, GetError(err, "SubmitJob", globalFailure)
 	}
 	if response != nil && !response.IsSuccess() {
 		c.metrics.submitJobFailureCounter.Inc(ctx)
 		logger.Warnf(ctx, fmt.Sprintf("Job submission failed with response %v", response))
-		return nil, errors.New(fmt.Sprintf("Job submission failed with status %v\n%s",
-			response.Status(), string(response.Body())))
+		return nil, GetError(err, "SubmitJob", response.Status())
```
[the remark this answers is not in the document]
glaksh100: This was fixed to include the response body in the error. E.g. event now:
```
Events:
  Type     Reason  Age   From              Message
  ----     ------  ----  ----              -------
  Normal   Create  39s   flinkk8soperator  Flink cluster created
  Warning  Create  24s   flinkk8soperator  Failed to submit job to cluster: SubmitJob call failed with status 400 Bad Request and message [{"errors":["Jar file /opt/flink/flink-web-upload/nonexistent-1.0.0-SNAPSHOT.jar does not exist"]}]
```

pkg/controller/flink/client/error_handler.go
```diff
@@ -0,0 +1,65 @@
...
+	"GetClusterOverview503":              {},
+	"GetClusterOverview" + globalFailure: {},
+	"CheckSavepointStatus500":            {},
+	"CheckSavepointStatus503":            {},
+	"CancelJobWithSavepoint500":          {},
+	"CancelJobWithSavepoint503":          {},
+}
+
+// FlinkApplicationError implements the error interface to make error handling more structured
+type FlinkApplicationError struct {
```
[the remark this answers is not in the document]
glaksh100: The error key is still a combination of the method and the status from the response. Do you think making the error verbose in the event and using only the key in the `LastSeenError` field would be sufficiently actionable for the users?

pkg/controller/flink/client/error_handler.go
glaksh100: Makes sense on both bounding and adding a jitter.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: Yeah I think it may make sense to add these two configs to the configmap? Does that make sense?

pkg/controller/flink/client/error_handler.go
glaksh100: Sounds good. I'll increase the number of retries here.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: Good point. This has been addressed in the `Handle()` loop. Let me know if that doesn't make sense.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: This has also been addressed, albeit some updates are required to the exponential backoff itself.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -23,6 +23,7 @@ import (
 
 const (
 	jobFinalizer = "job.finalizers.flink.k8s.io"
+	maxRetries   = 3
```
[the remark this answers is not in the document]
glaksh100: This is configurable per error now (all of them are set to a default, but the provision to tune it per error exists now).

pkg/controller/flink/client/error_handler.go
```diff
@@ -1,53 +1,59 @@
...
-var retryableErrors = map[string]struct{}{
-	"GetClusterOverview500":              {},
-	"GetClusterOverview503":              {},
-	"GetClusterOverview" + globalFailure: {},
-	"CheckSavepointStatus500":            {},
-	"CheckSavepointStatus503":            {},
-	"CancelJobWithSavepoint500":          {},
-	"CancelJobWithSavepoint503":          {},
+var retryableErrors = map[string]int32{
+	"GetClusterOverview500":              defaultRetries,
```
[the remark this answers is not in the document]
glaksh100: The error codes are still intended to be a combination of the method + the response status, given that the entire error string can be fairly verbose and non-deterministic. I definitely like the idea of the error codes being compiler checked. I'll address that.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: I've also included a time-based check on errors that are neither retryable nor deemed to fail fast. Let me know if that doesn't make sense.

pkg/controller/flink/client/error_handler.go
glaksh100: Enumerating all possible errors (method * httpErrorCodes) is perhaps going to be too long a list? My intention was to enumerate the errors that we know should be retryable or fail fast. On encountering errors that are of neither type, we default to waiting to see if the application makes progress for X minutes. I definitely get your point that we now have to remember to update these two maps. An alternative I can think of is to make the [GetError()](https://github.com/lyft/flinkk8soperator/blob/improve-staleness-logic/pkg/controller/flink/client/error_handler.go#L48) call to also update the map(s). Wdyt?

glaksh100:
Things I'm planning to address in the next pass:
1) Improving the exponential back off with jitter and a bound
2) Compile-check the error codes
3) Make the retryhandler parameters configurable and move them to the configMap

pkg/controller/flink/client/error_handler.go
glaksh100: The need for this should no longer exist as the maps have now been removed.

glaksh100:
Changelog:
1) Refactored the retryable/failfast errors to not depend on a map at all. When the error is logged, the error itself dictates what category it belongs to along with its `maxRetries`.
2) Increased `defaultRetries` and also added jitter to the retry delay calculation
3) Added retry configurations to the configmap (basebackoff/maxbackoff/maxwaitonerror)

pkg/controller/flink/client/error_handler.go
glaksh100: That's a great idea. I updated Handle() to handle the time to retry logic. Thanks for the suggestion.

glaksh100:
As of yesterday, I've addressed all comments afaict. PTAL when you have a chance @mwylde @anandswaminathan

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -83,16 +85,37 @@ func (s *FlinkStateMachine) shouldRollback(ctx context.Context, application *v1a
...
+
+	// For a class of errors that should fail fast, e.g. submitjob failures, always fail fast
+	if s.retryHandler.IsErrorFailFast(application.Status.LastSeenError) {
+		s.flinkController.LogEvent(ctx, application, "", corev1.EventTypeWarning, fmt.Sprintf("Application failed to progress in the %v phase with error %v", application.Status.Phase, application.Status.LastSeenError))
+		application.Status.LastSeenError = nil
+		return true
+	}
+
+	// All other errors, use a time based wait to determine whether to rollback or not.
+	if application.Status.LastSeenError != nil && application.Status.LastUpdatedAt != nil {
```
mwylde: I don't think this logic quite works for issues that arise during cluster start-up (for example, wrong image). The lastSeenError is never set (as the error occurs in the pod as is not visible to the operator).

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -124,27 +147,50 @@ func (s *FlinkStateMachine) handle(ctx context.Context, application *v1alpha1.Fl
...
+		switch application.Status.Phase {
+		case v1alpha1.FlinkApplicationNew, v1alpha1.FlinkApplicationUpdating:
+			// Currently just transitions to the next state
+			appErr = s.handleNewOrUpdating(ctx, application)
+		case v1alpha1.FlinkApplicationClusterStarting:
+			appErr = s.handleClusterStarting(ctx, application)
+		case v1alpha1.FlinkApplicationSubmittingJob:
+			appErr = s.handleSubmittingJob(ctx, application)
+		case v1alpha1.FlinkApplicationRunning, v1alpha1.FlinkApplicationDeployFailed:
+			appErr = s.handleApplicationRunning(ctx, application)
```
mwylde: I think we need different logic for the running phases. RIght now if we hit some issues during running (e.g., talking to the JM or k8s api server) we'll quickly exhaust all of our retries, and the operator will refuse to do anything until the application is deleted and recreated. If we're in running, we probably don't want to invoke any of the error handling stuff I think.

pkg/controller/flink/client/error_handler.go
```diff
@@ -0,0 +1,160 @@
...
+	GetCheckpointCounts    FlinkMethod = "GetCheckpointCounts"
+	GetJobOverview         FlinkMethod = "GetJobOverview"
+)
+
+// FlinkApplicationError implements the error interface to make error handling more structured
+type FlinkApplicationError struct {
+	AppError            string       `json:"appError,omitempty"`
+	Method              FlinkMethod  `json:"method,omitempty"`
+	ErrorCode           string       `json:"errorCode,omitempty"`
+	IsRetryable         bool         `json:"isRetryable,omitempty"`
```
mwylde: What's the difference between IsRetryable and IsFailFast?

pkg/controller/flink/client/error_handler.go
```diff
@@ -0,0 +1,160 @@
...
+func (f *FlinkApplicationError) DeepCopy() *FlinkApplicationError {
+	if f == nil {
+		return nil
+	}
+	out := new(FlinkApplicationError)
+	f.DeepCopyInto(out)
+	return out
+}
+
+func GetError(err error, method FlinkMethod, errorCode string, isRetryable bool, isFailFast bool, maxRetries int32, message ...string) error {
```
mwylde: I'd prefer if we can avoid having these two boolean arguments here. It's very hard to tell what they mean from reading function calls, and easy to mix them up.

Maybe instead have methods like GetRetryableError / GetFastFailError?

glaksh100: Fair point. Refactored to make these two separate methods.

pkg/controller/flink/client/error_handler.go
glaksh100: A fail fast error implies that the operator will move the application to a deployFailed state in a single attempt (for example if we encounter a 400 in a submit job request). The idea here is that we'll have knowledge of what errors are definitely retryable/failfast (akin to retryable vs. throwable exceptions), so we specify them explicitly and everything else is handled on a wait-based recovery.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: That makes sense to me. I've added a check to not invoke the error handling code when the application is in a running phase.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: Good point. Updated to have a generic time-based fallback for cases (with or without LastSeenErrors) that don't fall under retryable/fail fast categories.  Side note: Pod/cluster based errors are really hard to debug right now :(

pkg/controller/flink/client/api.go
```diff
@@ -217,18 +217,17 @@ func (c *FlinkJobManagerClient) SubmitJob(ctx context.Context, url string, jarID
...
 		c.metrics.submitJobFailureCounter.Inc(ctx)
-		return nil, errors.Wrap(err, "Submit job API request failed")
+		return nil, GetError(err, SubmitJob, GlobalFailure, false, true, noRetries)
 	}
 	if response != nil && !response.IsSuccess() {
 		c.metrics.submitJobFailureCounter.Inc(ctx)
 		logger.Warnf(ctx, fmt.Sprintf("Job submission failed with response %v", response))
-		return nil, errors.New(fmt.Sprintf("Job submission failed with status %v\n%s",
-			response.Status(), string(response.Body())))
+		return nil, GetError(err, SubmitJob, response.Status(), false, true, noRetries, string(response.Body()))
```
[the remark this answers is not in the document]
glaksh100: Thanks. Updated to retry on errors with > 499 status codes.

glaksh100:
PTAL @mwylde when you have a chance!

pkg/controller/flink/client/error_handler.go
mwylde: I still don't quite understand why we have both IsRetryable and IsFailFast. Like, given this truth table:

|   | IsRetryable | IsFailFast |
|---|-------------|------------|
| 1 | T           | T          |
| 2 | T           | F          |
| 3 | F           | T          |
| 4 | F           | F          |

Which of those rows makes sense? It seems to me that only rows 2 and 3 make sense, so we can reduce this to a single variable.

glaksh100: IMO 4) also makes sense, wherein when we encounter a class of errors that are [not](https://github.com/lyft/flinkk8soperator/blob/d4214227dd5a9c2b40eefb869e6f2510548e4b79/pkg/controller/flinkapplication/flink_state_machine.go#L636) necessarily from failing flink API calls (i.e. errors that we've not exhaustively enumerated), we default to a state of waiting for the application to make progress in the error-ed state. Alternatively, we could simplify all of this (as you suggested) and make aforementioned errors retryable by default.

mwylde: So in that case we're just waiting, but not retrying?

glaksh100: Yes. That's right.

glaksh100: Discussed offline and the approach we're going with is to only classify errors as retryable. Everything that's not retryable will be considered to "fail fast" i.e. move the application to a DeployFailed state in a single attempt.

glaksh100:
Thanks for the thorough review @mwylde!

PR #33, opened 2019-06-21
Adding openAPIV3Schema Validation for the Flink application custom resource

anandswaminathan:
Adding Validation of flink application custom objects using openAPIV3Schema. Helps fail the creation/update of custom resources with invalid format.

### DIFF
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -51,7 +51,7 @@ spec:
               enum: [Savepoint, None, ForceCancel]
             deploymentMode:
               type: string
-              enum: [Single, Dual]
+              enum: [Dual]
             rpcPort:
               type: integer
               minimum: 1
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -11,9 +11,357 @@ spec:
     singular: flinkapplication
     shortNames:
       - flinkapp
-
   scope: Namespaced
   version: v1alpha1
+  validation:
+    # openAPIV3Schema is the schema for validating custom objects.
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          properties:
+            image:
+              type: string
+            imagePullPolicy:
+              type: string
+              enum: [Always, Never, IfNotPresent]
+            imagePullSecrets:
+              type: array
+              items:
+                type: object
+                properties:
+                  name:
+                    type: string
+            jarName:
+              type: string
+            programArgs:
+              type: string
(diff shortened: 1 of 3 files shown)


deploy/crd.yaml
```diff
@@ -11,9 +11,357 @@ spec:
...
+              type: string
+            parallelism:
+              type: integer
+              minimum: 1
+            deleteMode:
+              type: string
+              enum: [Savepoint, None, ForceCancel]
+            deploymentMode:
+              type: string
+              enum: [Single, Dual]
```
mwylde: Since Single is not currently supported, should we remove?

deploy/crd.yaml
```diff
@@ -11,9 +11,357 @@ spec:
...
+              maximum: 65535
+            metricsQueryPort:
+              type: integer
+              minimum: 1
+              maximum: 65535
+            flinkConfig:
+              type: object
+              properties:
+                additionalProperties:
+                  type: string
```
mwylde: I think this is too restrictive. In the go definition this is `map[string]interface{}`, and there are flink configs that are numeric and boolean in addition to string.

deploy/crd.yaml
```diff
@@ -11,9 +11,357 @@ spec:
...
+                  type: integer
+                  minimum: 1
+                offHeapMemoryFraction:
+                  type: float
+                nodeSelector:
+                  type: object
+                  properties:
+                    additionalProperties:
+                      type: string
+                envConfig:
```
mwylde: Can we factor this out into its own object? That would avoid the need to duplicate it between TM/JM.

deploy/crd.yaml
anandswaminathan: Sure

deploy/crd.yaml
anandswaminathan: The problem with `map[string]interface{}` is that it can allow nested structure. On the other end we are appending only strings  to the conf.yaml.

I tested with integer values like `taskmanager.heap.size: 200` and boolean values and they work. It integer as numeric string and does not fail.

deploy/crd.yaml
anandswaminathan: I tried that I am unable to make components work for custom resources. Do you have any working example you can share ?

mwylde: Looks like it's not supported in kubernetes: https://github.com/kubernetes/kubernetes/issues/62872 😞

PR #37, opened 2019-06-26
Update logic around number and healthy task managers in status

anandswaminathan:
New Logic:
 - Uses the value of availableReplicas to update `NumberOfTaskManagers`
 - `HealthyTaskManagers` is still obtained from the API response
 - Health logic is now compared between `HealthyTaskManagers`, and number of replicas expected.

Cluster status is also updated when in `SubmittingJob` state.
Not updating the ClusterStatus in `ClusterStarting`, as there could be multiple clusters, and hence removed the check.

### DIFF
--- a/pkg/controller/flink/flink.go
+++ b/pkg/controller/flink/flink.go
@@ -152,6 +152,17 @@ func GetActiveFlinkJob(jobs []client.FlinkJob) *client.FlinkJob {
 	return nil
 }
 
+// returns the deployments for the application
+func (f *Controller) getDeploymentsForApplication(ctx context.Context, application *v1alpha1.FlinkApplication) (*v1.DeploymentList, error) {
+	labelMap := GetAppHashSelector(application)
+	depList, err := f.k8Cluster.GetDeploymentsWithLabel(ctx, application.Namespace, labelMap)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/flink/flink.go
```diff
@@ -442,36 +450,41 @@ func (f *Controller) LogEvent(ctx context.Context, app *v1alpha1.FlinkApplicatio
...
 	// Determine Health of the cluster.
-	// Error retrieving cluster / taskmanagers overview (after startup/readiness) --> Red
 	// Healthy TaskManagers == Number of taskmanagers --> Green
 	// Else --> Yellow
-	if clusterErrors != "" && (application.Status.Phase != v1alpha1.FlinkApplicationClusterStarting &&
-		application.Status.Phase != v1alpha1.FlinkApplicationSubmittingJob) {
-		application.Status.ClusterStatus.Health = v1alpha1.Red
-		return false, errors.New(clusterErrors)
-	} else if application.Status.ClusterStatus.HealthyTaskManagers == application.Status.ClusterStatus.NumberOfTaskManagers {
+	if application.Status.ClusterStatus.HealthyTaskManagers == tmDeployment.Status.Replicas {
```
glaksh100: Nice!

PR #38, opened 2019-07-02
Add support for non restored state on job submission

YuvalItzchakov:
This PR adds support for providing the `--allowNonRestoredState` flag to a submit job request. If, for some reason, that state isn't persistent between versions of a job, the deployment will fail entirely.

The addition here adds a `allowNonRestoredState` boolean flag to the `crd.yaml`.

### DIFF
--- a/pkg/controller/flink/flink.go
+++ b/pkg/controller/flink/flink.go
@@ -377,7 +377,7 @@ func (f *Controller) DeleteOldResourcesForApp(ctx context.Context, app *v1alpha1
 	for _, d := range deployments.Items {
 		if d.Labels[FlinkAppHash] != "" &&
 			d.Labels[FlinkAppHash] != curHash &&
-		// verify that this deployment matches the jobmanager or taskmanager naming format
+			// verify that this deployment matches the jobmanager or taskmanager naming format
 			(d.Name == fmt.Sprintf(JobManagerNameFormat, app.Name, d.Labels[FlinkAppHash]) ||
 				d.Name == fmt.Sprintf(TaskManagerNameFormat, app.Name, d.Labels[FlinkAppHash])) {
 			oldObjects = append(oldObjects, d.DeepCopy())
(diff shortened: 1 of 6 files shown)


pkg/controller/flink/flink.go
```diff
@@ -245,10 +245,11 @@ func (f *Controller) StartFlinkJob(ctx context.Context, application *v1alpha1.Fl
...
 		client.SubmitJobRequest{
-			Parallelism:   parallelism,
-			SavepointPath: application.Spec.SavepointInfo.SavepointLocation,
-			EntryClass:    entryClass,
-			ProgramArgs:   programArgs,
+			Parallelism:           parallelism,
+			SavepointPath:         application.Spec.SavepointInfo.SavepointLocation,
+			EntryClass:            entryClass,
+			ProgramArgs:           programArgs,
+			AllowNonRestoredState: application.Spec.AllowNonRestoredState,
```
anandswaminathan: Add to test that this field is populated.
https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flink/flink_test.go#L432

anandswaminathan:
Also you need to run `make generate`

YuvalItzchakov:
Appended a check to the existing test, created a new test to make sure `appendNonRestoredState` is propagated properly when set and ran `make generate`

anandswaminathan:
Great. Thanks @YuvalItzchakov

anandswaminathan:
@YuvalItzchakov This is not the right way to do it.

You will have to plumb the flag through the state machine. This case will not work when there is a rollback scenario.

Check this: https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L266

You will need to add the flag to the status as well.

YuvalItzchakov:
Thanks @anandswaminathan, I’ll get to work on the fix.

PR #39, opened 2019-07-02
Upgrade controller runtime and Multi Namespace support for Flink operator

anandswaminathan:
***[Backward Incompatible]***

The hash value computation has changed. Deploying this version of operator will cause rollover/restart of applications.

This PR upgrades the version of controller runtime. Uses a fixed version of Flytestdlib.

Flink operator will now support watching applications across multiple namespaces, previously it was either a single namespace or all namespaces.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -6,34 +6,37 @@ required = [
   "k8s.io/code-generator/cmd/client-gen",
   "k8s.io/code-generator/cmd/lister-gen",
   "k8s.io/code-generator/cmd/informer-gen",
-  "k8s.io/code-generator/cmd/openapi-gen",
   "k8s.io/gengo/args",
 ]
 
 [[constraint]]
-  name = "sigs.k8s.io/controller-runtime"
-  version = "^0.1.0"
-
-[[override]]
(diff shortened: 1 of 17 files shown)


Gopkg.toml
```diff
@@ -6,34 +6,37 @@ required = [
...
+  version = "kubernetes-1.14.1"
+
+[[override]]
+  name = "gopkg.in/fsnotify.v1"
+  source = "https://github.com/fsnotify/fsnotify.git"
 
 [[constraint]]
   name = "github.com/lyft/flytestdlib"
-  version = "0.2.8"
+  revision = "4e4da2f1ca1745127a79d299ef9048e86bd4e44c"
```
anandswaminathan: To be updated once I merge flytestdlib changes and generate a release

mwylde:
It's not clear to me why this PR (which seems limited to internal details of the operator) should cause a hash change. Can you expand on what's going on there?

anandswaminathan:
@mwylde The version of Controller Runtime that supports MultiNamespace, has constraint to use upgraded version of "k8s.io/api" - this updates the underlying Deployment object.

PR #41, opened 2019-07-08
Add more metrics to Flinkk8soperator

anandswaminathan:
- Add create/update/delete success and failure metrics for Kubernetes objects - in turn can help identify etcd issues/bugs.
- Metrics for deployments fetched from Cache/Client


glaksh100:
👍 LGTM!

PR #42, opened 2019-07-09
Find the correct deployments for readiness and status checks

mwylde:
We somehow ended up with two codepaths for finding the current deployments associated with an application. The one used by the status checks was using the current application hash, not the DeployHash, to find the deployments.. This caused it to fail for applications that had been rolled back with error logs like

```
ERRO[0000] Unable to find task manager deployment        app_name=operator-test-app ns=default phase=DeployFailed src="flink.go:461"
```

This PR removes that function and moves the callers to `GetCurrentDeploymentsForApp`, which also has the benefit of making the handling code a bit simpler.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -990,6 +990,7 @@
     "k8s.io/code-generator/cmd/informer-gen",
     "k8s.io/code-generator/cmd/lister-gen",
     "k8s.io/gengo/args",
+    "k8s.io/klog",
     "sigs.k8s.io/controller-runtime/pkg/cache",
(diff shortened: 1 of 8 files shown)


anandswaminathan:
+1

I see the issue, it is this check that was missing
```
	if appHash == app.Status.FailedDeployHash {
		return app.Status.DeployHash
	}
```

mwylde:
Yeah, that's what was missing. If we've failed to deploy, the current hash is the DeployHash, not Hash(app).

PR #44, opened 2019-07-11
[STRMCMP-558] Event improvements

mwylde:
This PR makes two improvements to the events that the operator emits for FlinkApplication resources.

First, it ensures that all deploy-related events are unique to a particular deploy (generally by including the hash of the resource in the message). This ensures that if two deploys occur in rapid succession, the second one will still produce events (if more than 10 events with the same message are emitted within a 10 minute period, later events are dropped). This prevents a situation where a developer does a deploy, encounters an error that produces several events, tries the deploy again but this time doesn't seen any events because we have already exhausted our limit.

The second change is to fix the reason field of the events we emit. When I originally added events I misunderstood what the reason field was. Re-reading the docs, I now understand it to be basically a machine-readable code for the event, with the message as its human-readable counterpart:

> 'reason' is the reason this event is generated. 'reason' should be short and unique; it should be in UpperCamelCase format (starting with a capital letter). "reason" will be used to automate handling of events, so imagine people writing switch statements to handle them. You want to make that easy.

Using unique reasons for each event also prevents distinct events from being improperly combined.

### DIFF
--- a/pkg/controller/flink/flink.go
+++ b/pkg/controller/flink/flink.go
@@ -82,7 +82,7 @@ type ControllerInterface interface {
 	FindExternalizedCheckpoint(ctx context.Context, application *v1alpha1.FlinkApplication, hash string) (string, error)
 
 	// Logs an event to the FlinkApplication resource and to the operator log
-	LogEvent(ctx context.Context, app *v1alpha1.FlinkApplication, eventType string, message string)
+	LogEvent(ctx context.Context, app *v1alpha1.FlinkApplication, eventType string, reason string, message string)
 
 	// Compares and updates new cluster status with current cluster status
 	// Returns true if there is a change in ClusterStatus
@@ -206,7 +206,7 @@ func (f *Controller) CreateCluster(ctx context.Context, application *v1alpha1.Fl
 	newlyCreatedJm, err := f.jobManager.CreateIfNotExist(ctx, application)
 	if err != nil {
 		logger.Errorf(ctx, "Job manager cluster creation did not succeed %v", err)
-		f.LogEvent(ctx, application, corev1.EventTypeWarning,
+		f.LogEvent(ctx, application, corev1.EventTypeWarning, "CreateClusterFailed",
 			fmt.Sprintf("Failed to create job managers for deploy %s: %v",
 				HashForApplication(application), err))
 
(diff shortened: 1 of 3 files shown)


pkg/controller/flink/flink.go
```diff
@@ -206,21 +206,24 @@ func (f *Controller) CreateCluster(ctx context.Context, application *v1alpha1.Fl
 	newlyCreatedJm, err := f.jobManager.CreateIfNotExist(ctx, application)
 	if err != nil {
 		logger.Errorf(ctx, "Job manager cluster creation did not succeed %v", err)
-		f.LogEvent(ctx, application, corev1.EventTypeWarning,
-			fmt.Sprintf("Failed to create job managers: %v", err))
+		f.LogEvent(ctx, application, corev1.EventTypeWarning, "CreateClusterFailed",
+			fmt.Sprintf("Failed to create job managers for deploy %s: %v",
+				HashForApplication(application), err))
```
glaksh100: HashForApplication() returns a different value in each of these calls (within the CreateCluster method i.e.)?

mwylde: It should be the same value. We're only ever executing one of these three calls, so there's no duplication of work.

anandswaminathan: Nit: hash := HashForApplication(app), and use hash.

pkg/controller/flink/flink.go
```diff
@@ -206,21 +206,24 @@ func (f *Controller) CreateCluster(ctx context.Context, application *v1alpha1.Fl
...
-			fmt.Sprintf("Failed to create task managers: %v", err))
+		f.LogEvent(ctx, application, corev1.EventTypeWarning, "CreateClusterFailed",
+			fmt.Sprintf("Failed to create task managers for deploy %s: %v",
+				HashForApplication(application), err))
 		return err
 	}
 
 	if newlyCreatedJm || newlyCreatedTm {
-		f.LogEvent(ctx, application, corev1.EventTypeNormal, "Flink cluster created")
+		f.LogEvent(ctx, application, corev1.EventTypeNormal, "CreatingCluster",
```
anandswaminathan: Nit, Move reason to constant variable ?

pkg/controller/flink/flink.go
```diff
@@ -388,7 +391,8 @@ func (f *Controller) DeleteOldResourcesForApp(ctx context.Context, app *v1alpha1
 	}
 
 	for k := range deletedHashes {
-		f.LogEvent(ctx, app, corev1.EventTypeNormal, fmt.Sprintf("Deleted old cluster with hash %s", k))
+		f.LogEvent(ctx, app, corev1.EventTypeNormal, "ToreDownCluster",
```
anandswaminathan: Same. Nit, Move reason to constant variable ?

mwylde: I don't really see any reason to. It's only used here and is descriptive of this event.

pkg/controller/flink/flink.go
glaksh100: Makes sense.

pkg/controller/flink/flink.go
anandswaminathan: Nothing major. Just a nit, so that they can be reused later.

anandswaminathan:
+1

mwylde:
🚀

PR #46, opened 2019-07-11
Add a ForceRollback option to CRD

glaksh100:
This PR adds the ability for a user to cancel an ongoing deploy/update. This is done in the following way:
- By adding a `CancelDeploy` to the Spec. The user sets this to be True to cancel the deploy. 
- When `CancelDeploy` is set to True, `shouldRollback` evaluates to True and moves the application to a `DeployFailed` state. 
- Either un-setting or setting `CancelDeploy` to False enables the deploy to be continued (as would be the case when we fix forward from DeployFailed)
- An integration test case is added to `simple_test.sh` for this workflow.
- CRD doc is updated to include `CancelDeploy`

Alternatively, I could have added an extra state (like `DeployCancelled`) but felt that that would have been quite a bit of code duplication as it'd be very similar to the DeployFailed state. Happy to incorporate other ideas!

### DIFF
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -106,5 +106,7 @@ Below is the list of fields in the custom resource and their description
   * **VolumeMounts** `type:[]v1.VolumeMount`
     Describes a mounting of a Volume within a container.
     
-  * **CancelDeploy** `type:bool`
-    Can be set to true to force cancel a deploy/update. Resetting to false or un-setting implies the deploy will continue.
+  * **ForceRollback** `type:bool`
+    Can be set to true to force rollback a deploy/update. The rollback is **not** performed when the application is in a **RUNNING** phase.
+    If an application is successfully rolled back, it is moved to a *DeployFailed* phase. Un-setting or setting `ForceRollback` to `False` will allow updates to progress normally.
+    
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -105,3 +105,6 @@ Below is the list of fields in the custom resource and their description
 
   * **VolumeMounts** `type:[]v1.VolumeMount`
     Describes a mounting of a Volume within a container.
+    
+  * **CancelDeploy** `type:bool`
+    Can be set to true to force cancel a deploy/update. Resetting to false or un-setting implies the deploy will continue.
--- a/integ/simple_test.go
+++ b/integ/simple_test.go
@@ -176,7 +176,6 @@ func (s *IntegSuite) TestSimple(c *C) {
 
 	{
 		newApp, err := s.Util.GetFlinkApplication(config.Name)
-		currHash := newApp.Status.DeployHash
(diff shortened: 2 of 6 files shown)


glaksh100:
cc. @mwylde / @anandswaminathan

docs/crd.md
```diff
@@ -105,3 +105,6 @@ Below is the list of fields in the custom resource and their description
 
   * **VolumeMounts** `type:[]v1.VolumeMount`
     Describes a mounting of a Volume within a container.
+    
+  * **CancelDeploy** `type:bool`
```
anandswaminathan: @glaksh100 

This is one of those use case that gets tricky in Kubernetes custom resource. You are looking to perform an action, when the underlying resource can be one of following states. This value is part of spec forever until the user changes it.

Let's say for a deploy from A -> B
App can be in A, or transitioning from A->B, or moved to B, while `CancelDeploy` is being set. 

Is your intention to tell the operator to preserve the application in State A ?

Wondering if renaming it `freezeDeploy` would be better, as the user has to come and set it to false explicitly for any future deploys to happen. Then we can store in the status the hash of the deploy to which the application is frozen.

Also this will only work if the app has not transitioned to B (and running). If not the application will be cancelled/frozen in B.

anandswaminathan: Also another option is `freezeDeployHash`, and the user has to provide the hash value.

In this case, we can store freezeStatus in the FlinkApplicationStatus, as true or false, if we were able to freeze at the right hash.

For the above example, if user says `freezeDeployHash=A`, but the app has already succeeded to B, then freezeStatus will be false. The problem here is this field `freezeDeployHash` will be unused and can live forever until reset by the user.

mwylde: I agree with Anand that CancelDeploy is a bit problematic. Not sure if freezeDeploy or freezeDeployHash are quite right either. The intention is not to freeze the deploy in place, but to force the deploy to fail and get rolled back.

Maybe RollbackHash, which when set to Status.DeployHash causes an immediately rollback to the previous version.

anandswaminathan: The RollbackHash the way you mentioned essentially results in a Freeze. All the future updates to the custom resource will not be triggered. 

Like I mentioned above, the only issue with have hash in spec is, by the time you set it, the deployhash underneath can change. Now we have a field in the spec that was never rolledback to and just confusing.

glaksh100:
Updates based on discussion (captured [here](https://docs.google.com/document/d/1cfuKVFLvwuBs6zvrQ1dXbVgB4XVXtNGRctQrnUQxlm8/edit))
- Renamed CRD to `forceRollback`
- Included a `rollbackHash` that will be set on rolling back the application
- Updated test and docs to reflect the change

cc @anandswaminathan / @mwylde

PR #48, opened 2019-07-15
Change error update logic and fix retry bug

glaksh100:
Currently, the `LastSeenError` is updated (on the k8s object) only if the error returned by a `handle<State>` call is not nil. This is a problem when `handle<State>` call succeeds after the first retry and `appErr` is now nil, the k8s object does not get updated and the error will continue to be retried until retries are exhausted. 

Changelog:
- Modify getAndUpdateError to compareAndUpdateError, which compares current and older error. If they are both nil, no updates to the k8s object are made. In all other cases, the k8s object is updated
- Call compareAndUpdateError irrespective of the value of appError
- Bug fix for the `IsTimeToRetry` logic which was evaluating a wrong value before and updated the corresponding test to not give false positives.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -25,7 +25,8 @@ import (
 
 const (
 	jobFinalizer         = "job.finalizers.flink.k8s.io"
-	shouldUpdateK8Object = true
+	applicationChanged   = true
+	applicationUnchanged = false
 )
 
 // The core state machine that manages Flink clusters and jobs. See docs/state_machine.md for a description of the
@@ -124,7 +125,16 @@ func (s *FlinkStateMachine) Handle(ctx context.Context, application *v1alpha1.Fl
 	successTimer := s.metrics.stateMachineHandleSuccessPhaseMap[currentPhase].Start(ctx)
 
 	defer timer.Stop()
-	err := s.handle(ctx, application)
+	updateApplication, err := s.handle(ctx, application)
+
+	// Update k8s object
+	if updateApplication {
+		updateAppErr := s.k8Cluster.UpdateK8Object(ctx, application)
+		if updateAppErr != nil {
+			s.metrics.errorCounterPhaseMap[currentPhase].Inc(ctx)
+			return updateAppErr
+		}
+	}
 	if err != nil {
 		s.metrics.errorCounterPhaseMap[currentPhase].Inc(ctx)
 	} else {
@@ -133,16 +143,16 @@ func (s *FlinkStateMachine) Handle(ctx context.Context, application *v1alpha1.Fl
 	return err
 }
 
-func (s *FlinkStateMachine) handle(ctx context.Context, application *v1alpha1.FlinkApplication) error {
+func (s *FlinkStateMachine) handle(ctx context.Context, application *v1alpha1.FlinkApplication) (bool, error) {
 	var appErr error
-	updateK8Object := !shouldUpdateK8Object
-	updateErrK8Object := !shouldUpdateK8Object
+	updateApplication := false
+	updateLastSeenError := false
 	appPhase := application.Status.Phase
 
 	if !application.ObjectMeta.DeletionTimestamp.IsZero() && application.Status.Phase != v1alpha1.FlinkApplicationDeleting {
 		// Always perform a single application update per callback
 		s.updateApplicationPhase(application, v1alpha1.FlinkApplicationDeleting)
-		updateK8Object = shouldUpdateK8Object
+		updateApplication = true
 	}
 
 	if !v1alpha1.IsRunningPhase(application.Status.Phase) {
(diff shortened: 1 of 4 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -162,16 +162,13 @@ func (s *FlinkStateMachine) handle(ctx context.Context, application *v1alpha1.Fl
 			appErr = s.handleApplicationDeleting(ctx, application)
 		}
 
-		if appErr != nil {
+		if !v1alpha1.IsRunningPhase(application.Status.Phase) {
```
anandswaminathan: You will have to cache phase as well in local variable.

phase := application.Status.Phase

So that you update for cases where you transition to `Running`

glaksh100: Thanks for mentioning, fixed!

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -648,7 +646,13 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
...
 
-func (s *FlinkStateMachine) getAndUpdateError(ctx context.Context, application *v1alpha1.FlinkApplication, err error) {
+func (s *FlinkStateMachine) compareAndUpdateError(ctx context.Context, application *v1alpha1.FlinkApplication, err error) {
+	oldErr := application.Status.LastSeenError
+
+	if err == nil && oldErr == nil {
+		return
+	}
+
 	if flinkAppError, ok := err.(*client.FlinkApplicationError); ok {
```
anandswaminathan: Wouldn't this break if err is nil ?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -648,7 +646,13 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
 	return nil
 }
 
-func (s *FlinkStateMachine) getAndUpdateError(ctx context.Context, application *v1alpha1.FlinkApplication, err error) {
+func (s *FlinkStateMachine) compareAndUpdateError(ctx context.Context, application *v1alpha1.FlinkApplication, err error) {
```
anandswaminathan: Hey,

Updating the status object in case of non error can fail here, if there has been an update in the handleLoop.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -648,7 +646,13 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
...
+	oldErr := application.Status.LastSeenError
+
+	if err == nil && oldErr == nil {
+		return
+	}
+
```
anandswaminathan: Can you update `GetRetryableError` to return `FlinkApplicationError` instead of error ?

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: Thanks for catching. Fixed.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: @anandswaminathan I've updated the logic now to minimize the number of updateK8Object call. Thanks for the suggestion!

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: I could. But I'm somewhat leaning towards not doing that; If the `api` returned FlinkApplicationError, which was then passed on to the controller (and then the state machine), we'd have to update all of the function signatures to return FlinkApplicationError(s).

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -386,17 +401,18 @@ func (s *FlinkStateMachine) updateGenericService(ctx context.Context, app *v1alp
...
+		s.updateApplicationPhase(app, v1alpha1.FlinkApplicationRollingBackJob)
+		return shouldUpdateK8Object, nil
 	}
 
 	// switch the service to point to the new jobmanager
 	hash := flink.HashForApplication(app)
 	err := s.updateGenericService(ctx, app, hash)
 	if err != nil {
-		return err
+		return !shouldUpdateK8Object, err
```
anandswaminathan: Hey `shouldUpdateK8Object` is applicable for different context.

Can you call it `applicationModified`, or `updateApplication`, `shouldUpdateApplication` ?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -24,7 +24,8 @@ import (
 )
 
 const (
-	jobFinalizer = "job.finalizers.flink.k8s.io"
+	jobFinalizer         = "job.finalizers.flink.k8s.io"
+	shouldUpdateK8Object = true
```
anandswaminathan: This should not be at the global level.

You should be created and returned within handle.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -134,44 +134,54 @@ func (s *FlinkStateMachine) Handle(ctx context.Context, application *v1alpha1.Fl
 }
 
 func (s *FlinkStateMachine) handle(ctx context.Context, application *v1alpha1.FlinkApplication) error {
+	var appErr error
+	updateK8Object := !shouldUpdateK8Object
```
anandswaminathan: You do not need `shouldUpdateK8Object`. Why not just initialize everything to false ?

anandswaminathan: Also you can simply make handle() return bool, error, and pass it to caller func `Handle` (notice caps)

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: This is only meant to be a const whose value is not updated. Instead of having true, false literals spread across handle methods, opted to use a constant that's a bit more explicitly named, but perhaps this is more confusing?

anandswaminathan: Oh I see. Yeah. this is confusing when you do `!`, without `!` this would have been clear.

Also lines https://github.com/lyft/flinkk8soperator/pull/48/files#diff-9d1b35757e5b357ae56fe201c420afb8R138 confused me. Wondering if this variable can be abused as we grow.

anandswaminathan: You can have another variable `applicationUnchanged` or `applicationModified` if you want to use it in Return statements.

glaksh100: Thanks for the feedback! Updated constants to `applicationChanged` and `applicationUnchanged` and limited their usage only in return statements.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -203,13 +216,14 @@ func (s *FlinkStateMachine) handleNewOrUpdating(ctx context.Context, application
 	err := s.flinkController.CreateCluster(ctx, application)
 	if err != nil {
 		logger.Errorf(ctx, "Cluster creation failed with error: %v", err)
-		return err
+		return !applicationChanged, err
```
anandswaminathan: Use applicationUnchanged ?

glaksh100: Sorry forgot to push the change :(

glaksh100:
@anandswaminathan thanks for identifying the bug and also for suggestions here! Merging now..

PR #49, opened 2019-07-18
Update to the latest Flinkk8soperator release

anandswaminathan:
We have changed the Docker file in our examples. Our current examples only work with the latest release of the operator (due to the latest Flink HA issue fix).

cc @mwylde @glaksh100


anandswaminathan:
Also cc @YuvalItzchakov @yuchaoran2011

anandswaminathan:
@glaksh100 That's a good point. 

Currently I am adding the change log in each release notes. But I believe once we switch to Beta, we will need to get that process going.

PR #50, opened 2019-07-22
Fix field names in json for CRD

anandswaminathan:
Noticed that some of the json field for certain attributes are different that intended. Fixing them here.


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -129,6 +126,8 @@ func (s *FlinkStateMachine) Handle(ctx context.Context, application *v1alpha1.Fl
 
 	// Update k8s object
 	if updateApplication {
+		now := v1.NewTime(s.clock.Now())
```
mwylde: This change seems unrelated?

anandswaminathan: We are currently not updating LastUpdatedAt when JobStatus/ClusterStatus is modified. 

This seems more logical, to set `LastUpdatedAt` before we make the update call. Happy to open in a different PR/

Issue #53, opened 2019-07-30
Out-of-date work count Docker image in Docker hub


jmdacruz:
The word count docker image has not been updated (~1 month ago) since #45 was merged (~12 days ago), and hence the quick start guide is broken (unless people build their own image against their own Docker registries, at which point the guide is no longer "quick" :-) ). Besides updating the example image in Docker hub, it would also be useful to tag it, so as to prevent this type of issues in the future.

anandswaminathan:
Thanks for pointing out.

I have built and pushed a new image now. https://cloud.docker.com/u/lyft/repository/registry-1.docker.io/lyft/wordcount-operator-example/tags

Will update the docs to remove usage of latest tag as well.

jmdacruz:
Sounds good! I just run the example using the tag for the image (`3b0347b2cdc1bda817e72b3099dac1c1b1363311`) and the cluster is now starting correctly.

PR #54, opened 2019-07-30
Fix image tag and update documentation

anandswaminathan:
This PR fixes use of latest tag in example wordcount. I have also update the quick start guide with latest CRD status and events output

Related issue: https://github.com/lyft/flinkk8soperator/issues/53


anandswaminathan:
cc @jmdacruz

PR #61, opened 2019-07-31
[Backward Incompatible] Modifying Hash function to be consistent

anandswaminathan:
Currently, whenever we upgrade "client-go", "apimachinery", "api"  the hash value also changes.

This PR changes the hashing mechanism

I have tested this across these two SHAs - `9aaa992bf12ecd0577a4b0f7dfb1c248a93d47ea` and `f9206646aaffb7763657ec9e794bc979b0b4722b`. The hash does not change.

### DIFF
--- a/pkg/controller/flink/container_utils.go
+++ b/pkg/controller/flink/container_utils.go
@@ -165,7 +165,7 @@ func HashForApplication(app *v1alpha1.FlinkApplication) string {
 	}
 
 	hasher := fnv.New32a()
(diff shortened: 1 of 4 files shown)


mwylde:
This change makes the hash sensitive to the order of fields in the json output, which are not guaranteed to be consistent across versions of the json library or go (Java 8 for example changed its hashmap implementation, which caused fields to get re-ordered compared to java 7).

Ideally we'd find a solution that is insensitive to details of the map or json implementations, while still handling missing fields correctly.

anandswaminathan:
@mwylde Updated to use Object hash  - https://github.com/benlaurie/objecthash

PR #73, opened 2019-08-13
Add Prometheus annotations

yuchaoran2011:
We've observed that by passing in the appropriate settings in `flink-conf.yaml` when building the Flink application image, Prometheus metrics can be properly exported from the JM and TM pods. But we've noticed that Prometheus is not able to scrape the exported metrics. The consequence is that from the Prometheus UI, the metrics won't show up and therefore no queryable. 

To let Prometheus discover exported metrics, the JM and TM pods need to have the right annotations, which are added in this PR. The default port is set to `9249`, which is the default value used in Flink documentation too.

### DIFF
--- a/pkg/controller/config/config.go
+++ b/pkg/controller/config/config.go
@@ -18,12 +18,13 @@ type Config struct {
 	ProfilerPort          config.Port     `json:"prof-port" pflag:"\"10254\",Profiler port"`
 	FlinkIngressURLFormat string          `json:"ingressUrlFormat"`
 	UseProxy              bool            `json:"useKubectlProxy"`
-	ProxyPort             config.Port     `json:"ProxyPort" pflag:"\"8001\",The port at which flink cluster runs locally"`
+	ProxyPort             config.Port     `json:"proxyPort" pflag:"\"8001\",The port at which flink cluster runs locally"`
 	ContainerNameFormat   string          `json:"containerNameFormat"`
 	Workers               int             `json:"workers" pflag:"4,Number of routines to process custom resource"`
 	BaseBackoffDuration   config.Duration `json:"baseBackoffDuration" pflag:"\"100ms\",Determines the base backoff for exponential retries."`
 	MaxBackoffDuration    config.Duration `json:"maxBackoffDuration" pflag:"\"30s\",Determines the max backoff for exponential retries."`
 	MaxErrDuration        config.Duration `json:"maxErrDuration" pflag:"\"5m\",Determines the max time to wait on errors."`
+	PrometheusPort        config.Port     `json:"prometheusPort" pflag:"\"9249\",The port at which Flink metrics are exported from job manager and task manager pods"`
 }
 
 func GetConfig() *Config {
--- a/pkg/controller/flink/container_utils.go
+++ b/pkg/controller/flink/container_utils.go
@@ -3,6 +3,7 @@ package flink
 import (
 	"fmt"
 	"hash/fnv"
+	"strconv"
 
 	"github.com/benlaurie/objecthash/go/objecthash"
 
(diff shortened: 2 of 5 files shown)


yuchaoran2011:
I realized that this change broke the test and API. I'm going to push a commit to fix them by making the new configurations optional.

pkg/controller/flink/container_utils.go
```diff
@@ -60,6 +61,13 @@ func getCommonAnnotations(app *v1beta1.FlinkApplication) map[string]string {
 	return annotations
 }
 
+func GetPrometheusAnnotations(port int) map[string]string {
+	return map[string]string{
+		"prometheus.io/scrape": "true",
```
jmdacruz: Shouldn't this be configurable too? I wouldn't return this annotation if `prometheus.io/scrape` is `false`.

yuchaoran2011: I'm going to and an optional configuration option called `enableMetrics` that has a default value of `false`. When it's false, no Prometheus-related annotations will be added at all.

pkg/controller/flink/job_manager_controller.go
```diff
@@ -307,7 +307,7 @@ func jobmanagerTemplate(app *v1beta1.FlinkApplication) *v1.Deployment {
 		ObjectMeta: metaV1.ObjectMeta{
 			Namespace:   app.Namespace,
 			Labels:      labels,
-			Annotations: getCommonAnnotations(app),
+			Annotations: GetPrometheusAnnotations(config.GetConfig().PrometheusPort.Port),
```
jmdacruz: Probably out of the scope of this particular PR, but considering that other (future) annotations could also be added to the JM and TM, it may be worthwhile considering creating a function that can render annotations from different sources (with the Prometheus annotations being one of them)

pkg/controller/flink/container_utils.go
jmdacruz: I'm wondering if it would be simpler to just have all annotations be explicitly listed on the configuration file (see my comment below on generic annotations). If you do this, then this would be extensible to any type of annotation, and not Prometheus-specific. Something like (potentially separating annotations for JM and TM):

```
operator:
   annotations:
      job_manager:
         prometheus.io/scrape: true
         prometheus.io/port: 9249
         foo: bar1
      task_manager:
         prometheus.io/scrape: true
         prometheus.io/port: 9249
         foo: bar2
```

jmdacruz: Actually, this seems related to the discussion going on in #10 (annotations on the ingress resource). Cc: @anandswaminathan

anandswaminathan: You can pass the annotations from the flink application right, like here - https://github.com/lyft/flinkk8soperator/blob/master/examples/wordcount/flink-operator-custom-resource.yaml#L6

pkg/controller/config/config.go
```diff
@@ -18,12 +18,13 @@ type Config struct {
...
 	FlinkIngressURLFormat string          `json:"ingressUrlFormat"`
 	UseProxy              bool            `json:"useKubectlProxy"`
-	ProxyPort             config.Port     `json:"ProxyPort" pflag:"\"8001\",The port at which flink cluster runs locally"`
+	ProxyPort             config.Port     `json:"proxyPort" pflag:"\"8001\",The port at which flink cluster runs locally"`
 	ContainerNameFormat   string          `json:"containerNameFormat"`
 	Workers               int             `json:"workers" pflag:"4,Number of routines to process custom resource"`
 	BaseBackoffDuration   config.Duration `json:"baseBackoffDuration" pflag:"\"100ms\",Determines the base backoff for exponential retries."`
 	MaxBackoffDuration    config.Duration `json:"maxBackoffDuration" pflag:"\"30s\",Determines the max backoff for exponential retries."`
 	MaxErrDuration        config.Duration `json:"maxErrDuration" pflag:"\"5m\",Determines the max time to wait on errors."`
+	PrometheusPort        config.Port     `json:"prometheusPort" pflag:"\"9249\",The port at which Flink metrics are exported from job manager and task manager pods"`
```
anandswaminathan: This would affect all the application in the cluster. Not a fan of setting configs at operator layer when it can be configured at individual application layer.

pkg/controller/flink/container_utils.go
yuchaoran2011: Oh I see. I totally missed that the user can configure annotations at the CR level. I'll close this PR then.

PR #74, opened 2019-08-13
Pass the Flink application annotations to Ingress and Service

anandswaminathan:
We currently only pass the annotations to JM/TM deployments. This change pass the annotations to Ingress and service as well.

### DIFF
--- a/pkg/controller/flink/ingress.go
+++ b/pkg/controller/flink/ingress.go
@@ -23,13 +23,14 @@ func GetFlinkUIIngressURL(jobName string) string {
 }
 
 func FetchJobManagerIngressCreateObj(app *flinkapp.FlinkApplication) *v1beta1.Ingress {
-	podLabels := common.DuplicateMap(app.Labels)
-	podLabels = common.CopyMap(podLabels, k8.GetAppLabel(app.Name))
+	ingressLabels := common.DuplicateMap(app.Labels)
+	ingressLabels = common.CopyMap(ingressLabels, k8.GetAppLabel(app.Name))
 
 	ingressMeta := v1.ObjectMeta{
-		Name:      app.Name,
-		Labels:    podLabels,
-		Namespace: app.Namespace,
+		Name:        app.Name,
+		Annotations: app.Annotations,
+		Labels:      ingressLabels,
+		Namespace:   app.Namespace,
 		OwnerReferences: []v1.OwnerReference{
 			*v1.NewControllerRef(app, app.GroupVersionKind()),
 		},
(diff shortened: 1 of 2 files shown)


jmdacruz:
Just one comment: since this could be used for Prometheus annotations (for metrics exposed by JM/TM), this means that if these same annotations are applied to the service then Prometheus would fail to scrape the endpoint for the service (this is because the way Prometheus uses annotations to discover metric targets). I guess the point is that it would be desirable to have different annotations depending on the type of object on the cluster.

anandswaminathan:
@jmdacruz Oh really. I assumed Prometheus does not scrape metrics from all kubernetes objects, but only from pods. 

I will hold on this, and think more. I have kind of restricted adding kubernetes types in custom resource as much as possible, as we might always change or modify the underlying implementation. Wondering if we can allow setting default labels/annotations for service/ingress/deployment through operator configuration.

jmdacruz:
@anandswaminathan yes, it's a bit confusing because of the way Kubernetes service discovery works in Prometheus, but you can scrape pretty much anything you want (see [here](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L135), including pods, services, ingress, etc.). I do agree with your design goal of not adding Kubernetes types in the CRD, and I believe I made a similar recommendation of temporarily using the operator configuration [here](https://github.com/lyft/flinkk8soperator/pull/73#discussion_r313469133) on #73 (now closed), except I was suggesting adding annotations directly for JM/TM, but I guess you could do the same thing for service/ingress/deployment

mwylde:
I agree with @anandswaminathan that we don't necessarily want to go down the road of making all details of the operator's resource creation configurable in the custom resource. It's unlikely that we can satisfy all the possible use cases without adding a lot of complexity to the resource (e.g., how could we support adding labels to just jobmanager pods? annotations to just the versioned service? etc.).

You may want to look into using a [mutating admission webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/), which should allow you to add your own annotations to the objects with whatever logic you need.

jmdacruz:
@mwylde Agree, mutating admission webhooks look like the ideal solution. Are annotations from the FlinkApplication already passed down to resources created by the operator? This would allow the mutating admission webhook to use the metadata for mutating the resources (e.g., use the Prometheus port indicated by the FlinkApplication instead of hardcoding it)

mwylde:
Currently annotations from the FlinkApplication are set on deployments and pods, but not ingresses or services.

PR #81, opened 2019-08-20
Add Web UI information to status

glaksh100:
This PR adds the cluster and job overview URLs to the cluster and job status respectively. The objective is to provide users easy access to the required Web UI. This information can also be used in any other CI/D tooling used to support Flink applications on k8s.  

Example output:
```yaml
Status:
  Cluster Status:
    Available Task Slots:     1
    Cluster Overview URL:     http://localhost:8001/api/v1/namespaces/default/services/simpleflinkk8s:8081/proxy/#/overview
    Health:                   Green
    Healthy Task Managers:    1
    Number Of Task Managers:  1
    Number Of Task Slots:     2
  Deploy Hash:                18051096
  Failed Deploy Hash:
  Job Status:
    Completed Checkpoint Count:  10
    Entry Class:                 com.lyft.OperatorTestApp
    Failed Checkpoint Count:     0
    Health:                      Green
    Jar Name:                    operator-test-app-1.0.0-SNAPSHOT.jar
    Job ID:                      c669950ee11c64980180eeaf9e62a63f
    Job Overview URL:            http://localhost:8001/api/v1/namespaces/default/services/simpleflinkk8s:8081/proxy/#/jobs/c669950ee11c64980180eeaf9e62a63f
    Job Restart Count:           0
```

### DIFF
--- a/pkg/apis/app/v1alpha1/types.go
+++ b/pkg/apis/app/v1alpha1/types.go
@@ -127,7 +127,6 @@ type SavepointInfo struct {
 }
 
 type FlinkClusterStatus struct {
-	ClusterOverviewURL   string       `json:"clusterOverviewURL,omitEmpty"`
 	Health               HealthStatus `json:"health,omitempty"`
 	NumberOfTaskManagers int32        `json:"numberOfTaskManagers,omitempty"`
 	HealthyTaskManagers  int32        `json:"healthyTaskManagers,omitepty"`
@@ -136,10 +135,9 @@ type FlinkClusterStatus struct {
 }
 
 type FlinkJobStatus struct {
-	JobOverviewURL string       `json:"jobOverviewURL,omitEmpty"`
-	JobID          string       `json:"jobID,omitEmpty"`
-	Health         HealthStatus `json:"health,omitEmpty"`
-	State          JobState     `json:"state,omitEmpty"`
+	JobID  string       `json:"jobID,omitEmpty"`
+	Health HealthStatus `json:"health,omitEmpty"`
+	State  JobState     `json:"state,omitEmpty"`
 
 	JarName               string `json:"jarName"`
 	Parallelism           int32  `json:"parallelism"`
--- a/pkg/apis/app/v1alpha1/types.go
+++ b/pkg/apis/app/v1alpha1/types.go
@@ -127,6 +127,7 @@ type SavepointInfo struct {
 }
 
 type FlinkClusterStatus struct {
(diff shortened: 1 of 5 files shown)


pkg/apis/app/v1alpha1/types.go
```diff
@@ -135,9 +136,10 @@ type FlinkClusterStatus struct {
 }
 
 type FlinkJobStatus struct {
-	JobID  string       `json:"jobID,omitEmpty"`
-	Health HealthStatus `json:"health,omitEmpty"`
-	State  JobState     `json:"state,omitEmpty"`
+	JobOverviewURL string       `json:"jobOverviewURL,omitEmpty"`
+	JobID          string       `json:"jobID,omitEmpty"`
+	Health         HealthStatus `json:"health,omitEmpty"`
+	State          JobState     `json:"state,omitEmpty"`
```
glaksh100: Not entirely sure if I should be updating v1alpha1 anymore?

pkg/controller/flink/flink.go
```diff
@@ -140,6 +140,14 @@ func getURLFromApp(application *v1beta1.FlinkApplication, hash string) string {
 	return fmt.Sprintf("http://%s.%s:%d", service, application.Namespace, port)
 }
 
+func getGenericURLFromApp(application *v1beta1.FlinkApplication) string {
```
anandswaminathan: This seems to be the `service` url, and is not exposed outside of the kubernetes cluster. Is that the intention ?

glaksh100: Hmm, this seems to be exposed in the local environment. i.e. for example for a locally running app — http://localhost:8001/api/v1/namespaces/default/services/simpleflinkk8s:8081/proxy/#/overview is exposed. But maybe works differently in a non-local environment?

pkg/apis/app/v1alpha1/types.go
mwylde: I think we should leave v1alpha1 as is going forward.

pkg/controller/flink/flink.go
mwylde: Yeah, this should be returning the external URL, which for non-proxy setups is defined by the ingress.

E.g. for our clusters it would look like, `<appname>.ingress.core-staging-4.us-east-1.k8s.lyft.net`

glaksh100: Okay makes sense. I updated this for non-proxy setups to return `GetFlinkUIIngressURL`.

glaksh100:
cc. @anandswaminathan @mwylde Updated this to store the ingress URL. PTAL when you have a chance.

Issue #76, opened 2019-08-15
Support specifying a service account for JM and TM deployments


yuchaoran2011:
The Flink operator deployment itself has a configurable service account. But the generated JM and TM deployments both run using an unspecified (i.e. `default`) service account. It will be useful to allow the user to specify a service account to be used for JM and TMs.

PR #83, opened 2019-08-21
Added an optional config in the CR to specify Flink application service account

yuchaoran2011:
#76

### DIFF
--- a/pkg/apis/app/v1beta1/types.go
+++ b/pkg/apis/app/v1beta1/types.go
@@ -106,6 +106,7 @@ type JobManagerConfig struct {
 	Replicas              *int32                      `json:"replicas,omitempty"`
 	OffHeapMemoryFraction *float64                    `json:"offHeapMemoryFraction,omitempty"`
 	NodeSelector          map[string]string           `json:"nodeSelector,omitempty"`
+	ServiceAccount        string                      `json:"serviceAccount,omitempty"`
 }
 
 type TaskManagerConfig struct {
@@ -114,6 +115,7 @@ type TaskManagerConfig struct {
 	TaskSlots             *int32                      `json:"taskSlots,omitempty"`
 	OffHeapMemoryFraction *float64                    `json:"offHeapMemoryFraction,omitempty"`
 	NodeSelector          map[string]string           `json:"nodeSelector,omitempty"`
+	ServiceAccount        string                      `json:"serviceAccount,omitempty"`
 }
 
 type EnvironmentConfig struct {
--- a/pkg/apis/app/v1beta1/types.go
+++ b/pkg/apis/app/v1beta1/types.go
@@ -33,7 +33,7 @@ type FlinkApplicationSpec struct {
 	Image                 string                       `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
 	ImagePullPolicy       apiv1.PullPolicy             `json:"imagePullPolicy,omitempty" protobuf:"bytes,14,opt,name=imagePullPolicy,casttype=PullPolicy"`
 	ImagePullSecrets      []apiv1.LocalObjectReference `json:"imagePullSecrets,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,15,rep,name=imagePullSecrets"`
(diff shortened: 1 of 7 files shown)


yuchaoran2011:
Looks like the unit test failed because the hash changed. But adding an optional field in the CR with a default value isn't an API-breaking change, or is it?

anandswaminathan:
Currently it is. To make it non breaking, you will have to set serviceAccount only if the serviceAccount is set in the flink application.

yuchaoran2011:
Got it. Now the tests are fixed as suggested.

mwylde:
Hey @yuchaoran2011 — is this something you could set using a mutating admission controller webhook? I'm a bit leery of going down the road of making every deployment / pod / service option configurable via the FlinkApplication resource.

yuchaoran2011:
I haven't tried but that sounds like much extra work to satisfy a use case that is as common as setting the service account. And I'm not "making every deployment/pod/service" configurable, just the **service accounts of the application JM and TM deployments**, not the operator deployment and not any other resource kind.

anandswaminathan:
Agree with @yuchaoran2011. ‘ServiceAccount’ is similar to other fields we already have in the custom resource like volume, env, ImagePullSecrets, etc. It fits in well with general kubernetes pattern. Every resource has a serviceAccount. Flink application is just another resource like a deployment or cron.

pkg/apis/app/v1beta1/types.go
```diff
@@ -33,6 +33,7 @@ type FlinkApplicationSpec struct {
 	Image                 string                       `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
 	ImagePullPolicy       apiv1.PullPolicy             `json:"imagePullPolicy,omitempty" protobuf:"bytes,14,opt,name=imagePullPolicy,casttype=PullPolicy"`
 	ImagePullSecrets      []apiv1.LocalObjectReference `json:"imagePullSecrets,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,15,rep,name=imagePullSecrets"`
+	ServiceAccount        string                       `json:"serviceAccount,omitempty"`
```
anandswaminathan: Can we call it serviceAccountName - consistent with other kubernetes resources?

pkg/controller/flink/config.go
```diff
@@ -42,6 +43,10 @@ func getJobmanagerReplicas(app *v1beta1.FlinkApplication) int32 {
 	return firstNonNil(app.Spec.JobManagerConfig.Replicas, JobManagerDefaultReplicaCount)
 }
 
+func getFlinkSa(app *v1beta1.FlinkApplication) string {
+	return app.Spec.ServiceAccount
```
anandswaminathan: Rename to getServiceAccountName ?

pkg/apis/app/v1beta1/types.go
yuchaoran2011: Done.

yuchaoran2011:
The integration test failed with error "the job exceeded maximum log depth". Not sure what happened. If one of the admins can restart the integration test, maybe the error will be gone.

mwylde:
Thanks @anandswaminathan for the explanation. That sounds reasonable.

@yuchaoran2011 unfortunately, that's just flakiness in the integration tests. It should pass if you re-run it (and we need to spend some time getting the tests reliable again).

yuchaoran2011:
Yeah I guessed so. But I don't think I have the permission to re-run the tests. If any one of you could do it, that would be great.

pkg/controller/flink/job_manager_controller.go
```diff
@@ -335,6 +335,12 @@ func jobmanagerTemplate(app *v1beta1.FlinkApplication) *v1.Deployment {
 			},
 		},
 	}
+
+	flinkSa := getServiceAccountName(app)
```
anandswaminathan: Nit: rename flinkSa --> serviceAccountName

mwylde:
I've restarted the job, but it's unfortunate that you weren't able to yourself. I'll look into getting that fixed.

anandswaminathan:
@yuchaoran2011 Merging this.

PR #87, opened 2019-08-22
[STRMCMP-616] Use error retrying system for savepoint errors during deletion

mwylde:
Currently, if a flinkapplication is being deleted with `DeleteMode: Savepoint` and the savepoint operation is failing, we will retry endlessly with no delay on the savepoint operation. This is because in that case, we don't return an error back to the error management system of the state machine.

This PR fixes that by properly returning an error for that situation. This causes us to do retry with exponential backoff instead of on every statemachine iteration.

It also should fix the "exceeded the maximum log length" integration test failures, which are caused by endless logging of the retries.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -2,6 +2,7 @@ package flinkapplication
 
 import (
 	"context"
+	"math"
 	"time"
 
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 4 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -676,6 +677,8 @@ func (s *FlinkStateMachine) handleApplicationDeleting(ctx context.Context, app *
 					fmt.Sprintf("Failed to take savepoint %v", status.Operation.FailureCause))
 				// clear the trigger id so that we can try again
 				app.Spec.SavepointInfo.TriggerID = ""
+				return true, client.GetRetryableError(errors.New("failed to take savepoint"),
```
anandswaminathan: Add a unit test ?

mwylde:
Updated with unit test and a config change to increase the maximum retry delay when running integration tests. This reduces the number of log lines in the integration test by ~1k.

PR #98, opened 2019-09-05
Add The Trade Desk to list of users

JonnyIncognito:


JonnyIncognito:
> CLA check — An author in one of the commits has no associated github name

I subsequently added the e-mail address used in the commit to my GitHub account, so I imagine this would pass now

PR #105, opened 2019-09-19
Minor updates

yuchaoran2011:
* Updated Go to 1.13
* Updated controller-runtime to 0.2.2
* Added Lightbend to who-is-using


yuchaoran2011:
I don't think the integration test failure is relevant to the PR. Could one of the admins trigger a re-run?

Issue #97, opened 2019-09-04
Fractional {job,task}manager.heap.size crashes Flink


JonnyIncognito:
The offHeapMemoryFraction can cause fractional values to be written to flink-conf.yaml, which crashes Flink on startup:

```
/opt/flink/bin/config.sh: line 261: [ERROR] >> 20: syntax error: operand expected (error token is "[ERROR] >> 20")
[ERROR] Configured JobManager memory size is not a valid value. Please set 'jobmanager.heap.size' in flink-conf.yaml.
```

For example, the following FlinkApplication...

```
  jobManagerConfig:
    resources:
      requests:
        memory: "4Gi"
        cpu: "1.0"
    offHeapMemoryFraction: 0.15
    replicas: 1
  taskManagerConfig:
    taskSlots: 1
    resources:
      requests:
        memory: "4Gi"
        cpu: "1.0"
    offHeapMemoryFraction: 0.3
```

... causes the following to be generated:

```
jobmanager.heap.size: 3481.6
taskmanager.heap.size: 2867.2
```

Above is using v0.2.0

anandswaminathan:
@JonnyIncognito 

Aah yes, it is is float https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flink/config.go#L87

Want to contribute a fix ?

JonnyIncognito:
Sure! I'm a Go newbie but seems like a nice simple task to start with.

anandswaminathan:
Also we have a slack channel - [slack](https://join.slack.com/t/flinkk8soperator/shared_invite/enQtNzIxMjc5NDYxODkxLTEwMThmN2I0M2QwYjM3ZDljYTFhMGRiNDUzM2FjZGYzNTRjYWNmYTE1NzNlNWM2YWM5NzNiNGFhMTkxZjA4OGU)

Feel free to join.

JonnyIncognito:
Addition to this: I noticed that the memory characteristics weren't correct in our metrics, which eventually led me to realise that the two config parameters are using the wrong units. Same generated example as above:

```
jobmanager.heap.size: 3481.6
taskmanager.heap.size: 2867.2
```

[...]

PR #106, opened 2019-09-25
[#97] [STRMCMP-648] [BREAKING] Fix rendering of flink memory configs

mwylde:
**Note that this is a breaking change, and will cause redeploy of all flink applications**

This PR fixes how we render Flink heap settings in the config based on the container memory requests and the configured `offHeapMemoryFraction`. This addresses #97, but also fixes a deeper issue where our memory settings were just being ignored.

This is a breaking change which is unfortunate, but also unavoidable in this case, as the current configuration is broken.

Currently we compute the Flink heap settings as `(1 - offHeapMemoryFraction) * requestedMemory`, convert it as a float to MB, then render it in the config like

```
jobmanager.heap.size: 1024
```

However this the wrong format and is ignored, at least by Flink 1.8; it should actually be 

```
jobmanager.heap.size: 1024m
````

In the case of fractional MBs, we print it as, say

```
jobmanager.heap.size: 1024.4
```

which causes startup to crash, as reported in #97. 

This PR fixes both issues by using KBs instead of MBs to give more granularity, rounding to the nearest KB, and including the unit:

```
jobmanager.heap.size: 102440k
```

### DIFF
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -97,6 +97,8 @@ spec:
                   minimum: 1
                 offHeapMemoryFraction:
                   type: number
+                  minimum: 0
+                  maximum: 1
                 nodeSelector:
                   type: object
                   properties:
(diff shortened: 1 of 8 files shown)


glaksh100:
👍 LGTM.

PR #114, opened 2019-10-04
Update slack link

mwylde:
Updates the slack link in both places in the readme to a shortlink (http://go.lyft.com/flinkoperator_slack) that we can always keep up to date.


anandswaminathan:
Nice!

PR #117, opened 2019-10-09
[STRMCMP-639] Trigger immediate rollback when job submission fails

mwylde:
This PR adds another case to our immediate rollback (i.e., unretryable error) logic, that catches the case where the job submission fails. This typically happens due to an incompatible savepoint or an invalid jobgraph.

### DIFF
--- a/pkg/controller/flink/client/api.go
+++ b/pkg/controller/flink/client/api.go
@@ -38,6 +38,13 @@ const retryCount = 3
 const httpGetTimeOut = 5 * time.Second
 const defaultTimeOut = 1 * time.Minute
 
+// ProgramInvocationException is thrown when the entry class doesn't exist or throws an exception
+const programInvocationException = "org.apache.flink.client.program.ProgramInvocationException"
+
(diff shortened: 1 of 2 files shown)


pkg/controller/flink/client/api.go
```diff
@@ -231,7 +231,11 @@ func (c *FlinkJobManagerClient) SubmitJob(ctx context.Context, url string, jarID
 			// Flink returns a 500 when the entry class doesn't exist or crashes on start, but we want to fail fast
 			// in those cases
 			body := response.String()
-			if strings.Contains(body, "org.apache.flink.client.program.ProgramInvocationException") {
+			// ProgramInvocationException is thrown when the entry class doesn't exist or throws an exception
+			if strings.Contains(body, "org.apache.flink.client.program.ProgramInvocationException") ||
```
anandswaminathan: Nit: Can you move "org.apache.flink.client.program.ProgramInvocationException" to a const err variable ?

mwylde: sure

PR #142, opened 2019-12-09
Fixes to ensure Delete phase is always handled and prevent deploys from being stuck in a Savepointing phase

glaksh100:
**Problem** 
- Deploy starts — Savepoint is triggered and triggerID registered
- `GetSavepointStatus` call keeps failing for whatever reason
-  Retries are exhausted for `GetSavepointStatus`, however we have[ no real way to progress ](https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L315)after this. The app gets stuck in a Savepointing Phase.
- User tries to force delete the application. However, with retry counts not being reset and a non-nil LastSeenError, the operator never handles the Deleting phase (i.e. [isTimeToHandlePhase](https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L196) always evaluates to false).

The PR fixes these problems in the following way:
- Reduces retries for the GetSavepointStatus call from 20 --> 3 to reduce the time to detect savepoint failures (esp. 404s around savepoints not being found for a triggerID). I considered making this non-retryable, but I'm guessing there's a possibility that the savepoint status for a triggerID may not become immediately available. 
- When retries for GetSavepointStatus are exhausted, we cannot reliably rollback at this point, as we may not have a running job or a successful savepoint. So we allow for the remainder of the savepointing block to complete i.e. try to recover from an externalized checkpoint. 
- When a Delete is attempted, always handle the Deleting phase irrespective of what error/retry state the application is in.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -336,13 +336,15 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
 	}
 
 	var restorePath string
-	if savepointStatusResponse != nil && savepointStatusResponse.Operation.Location == "" &&
-		savepointStatusResponse.SavepointStatus.Status != client.SavePointInProgress {
+	if rollback || (savepointStatusResponse.Operation.Location == "" &&
+		savepointStatusResponse.SavepointStatus.Status != client.SavePointInProgress) {
 		// Savepointing failed
 		// TODO: we should probably retry this a few times before failing
-		s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "SavepointFailed",
-			fmt.Sprintf("Failed to take savepoint for job %s: %v",
-				application.Status.JobStatus.JobID, savepointStatusResponse.Operation.FailureCause))
+		if savepointStatusResponse != nil {
+			s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "SavepointFailed",
+				fmt.Sprintf("Failed to take savepoint for job %s: %v",
+					application.Status.JobStatus.JobID, savepointStatusResponse.Operation.FailureCause))
+		}
 
 		// try to find an externalized checkpoint
 		path, err := s.flinkController.FindExternalizedCheckpoint(ctx, application, application.Status.DeployHash)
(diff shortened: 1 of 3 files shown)


glaksh100:
@mwylde can you PTAL?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -312,12 +319,22 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
...
+	// The application is NOT actually rolled back  because we're in a spot where there may be no application running and we don't
+	// have information on the last successful savepoint.
+	// In the event that rollback evaluates to True, we don't return immediately but allow for the remaining steps to proceed as normal.
+	rollback, reason := s.shouldRollback(ctx, application)
+
+	if err != nil && !rollback {
 		return statusUnchanged, err
 	}
 
+	s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "SavepointStatusCheckFailed",
```
mwylde: It looks like we'll print this out every time we savepoint, regardless of whether this failed

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -312,12 +319,22 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
 
 	// check the savepoints in progress
 	savepointStatusResponse, err := s.flinkController.GetSavepointStatus(ctx, application, application.Status.DeployHash)
-	if err != nil {
+	// Here shouldRollback() is used as a proxy to identify if we have exhausted all retries trying to GetSavepointStatus
+	// The application is NOT actually rolled back  because we're in a spot where there may be no application running and we don't
+	// have information on the last successful savepoint.
+	// In the event that rollback evaluates to True, we don't return immediately but allow for the remaining steps to proceed as normal.
```
mwylde: I think that if we fail to get the savepoint status, we will still not be able to recover. The following steps (checking for externalized checkpoint etc.) still require that we have a non-nil savepointStatusResponse.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: My bad. Yes. Fixed.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: True. This was a bad assumption (that we get return the response as is with the error), reinforced by bad unit tests. Fixed both, thanks for pointing out.

PR #144, opened 2019-12-12
Gate statemachine progress on taskmanager availability

mwylde:
Currently, the only condition for moving from ClusterStarting to Savepointing (at which point the job goes down) is that all JM/TM pods are up according to the deployment. However, various issues with the TM process or configuration can prevent them from actually registering with the JobManager and becoming available to run tasks. This can lead to extended downtime and can require manual intervention to fix. I've also added a check from the SubmittingJob -> Running transition that the tasks are actually running, which gives us a chance to automatically roll back if the job never successfully starts.

This PR also adds some more visibility into task-level status, so that users can tell if the job is really running (in Flink, a job can be in the Running state even if none of its tasks are running). I've added two new fields to the JobStatus (TotalTasks and RunningTasks) and updated the JobHealth logic to take into account whether tasks are actually running.

### DIFF
--- a/pkg/apis/app/v1beta1/types.go
+++ b/pkg/apis/app/v1beta1/types.go
@@ -155,6 +155,9 @@ type FlinkJobStatus struct {
 	RestorePath              string       `json:"restorePath,omitempty"`
 	RestoreTime              *metav1.Time `json:"restoreTime,omitempty"`
 	LastFailingTime          *metav1.Time `json:"lastFailingTime,omitempty"`
+
+	RunningTasks int32 `json:"runningTasks,omitempty"`
+	TotalTasks   int32 `json:"totalTasks,omitempty"`
 }
 
 type FlinkApplicationStatus struct {
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -271,14 +271,13 @@ func (s *FlinkStateMachine) handleClusterStarting(ctx context.Context, applicati
 
 	// Wait for all to be running
 	clusterReady, err := s.flinkController.IsClusterReady(ctx, application)
-	if err != nil {
+	if err != nil || !clusterReady {
 		return statusUnchanged, err
 	}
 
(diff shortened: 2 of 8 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -270,11 +270,15 @@ func (s *FlinkStateMachine) handleClusterStarting(ctx context.Context, applicati
...
 	// Wait for all to be running
-	ready, err := s.flinkController.IsClusterReady(ctx, application)
+	clusterReady, err := s.flinkController.IsClusterReady(ctx, application)
 	if err != nil {
 		return statusUnchanged, err
 	}
-	if !ready {
+
+	// ignore the error, we just care whether it's ready or not
+	serviceReady, _ := s.flinkController.IsServiceReady(ctx, application, flink.HashForApplication(application))
```
anandswaminathan: Since you do this here, you do not need to check `IsServiceReady` in SubmittingJob state right?

mwylde: Yeah, that check is probably not necessary any more

mwylde: Removed the check in SubmittingJob

pkg/apis/app/v1beta1/types.go
```diff
@@ -155,6 +155,9 @@ type FlinkJobStatus struct {
 	RestorePath              string       `json:"restorePath,omitempty"`
 	RestoreTime              *metav1.Time `json:"restoreTime,omitempty"`
 	LastFailingTime          *metav1.Time `json:"lastFailingTime,omitempty"`
+
+	RunningTasks int32 `json:"runningTasks,omitempty"`
+	TotalTasks   int32 `json:"totalTasks,omitempty"`
```
anandswaminathan: gofmt?

mwylde: This is what `golangci-lint --fix` produces. I think it only aligns stuff if there is no line break (see other examples in this file for similar formatting).

PR #145, opened 2019-12-13
Various minor fixes

yuchaoran2011:
* `omitempty` is the right tag name. Note the lowercase `e`.
* config strings should be camel case, with the first letter lowercase.
* Updated the example wordcount to use Flink 1.9.1
* Go version update to 1.13.4.


yuchaoran2011:
I wonder why the lint job failed because `make lint` ran without errors on my local machine. I'm using Go 1.13.4.  Is it a transient error, in which case could an admin trigger a re-run? Or is it something wrong with Travis? Let me know if it's something in the PR that I can fix. Thanks!

skonto:
This requires the new version for golangci-lint/ as described[ here](https://github.com/golangci/golangci-lint/issues/663#issuecomment-529871239).

yuchaoran2011:
Admins, as @skonto pointed out above, `golangci-lint` in travis is outdated. Do you think you guys can easily upgrade that? If not, I'll revert the upgrade to Go 1.13.4 to make the build pass.

mwylde:
Yeah, for now I would revert the change to 1.13. I'll look into what's required to update golangci-lint.

yuchaoran2011:
@mwylde Thanks! Done. The build is now passing.

PR #149, opened 2019-12-21
Provide operator configuration via FLINK_PROPERTIES fixes #135

### LINKED ISSUE #135: Replace environment variable for configuration injection
Environment variable `OPERATOR_FLINK_CONFIG` should be replaced with `FLINK_PROPERTIES` to support the stock Flink Docker images.

https://github.com/docker-flink/docker-flink/pull/82

We should also update the example to use that image.

tweise:
The old environment variable is retained for a transition period. The docker-flink containers now support environment variable substitution as well, so we can get rid of the custom entry point!


pkg/controller/flink/task_manager_controller_test.go
```diff
@@ -136,7 +136,7 @@ func TestTaskManagerHACreateSuccess(t *testing.T) {
 			"high-availability.cluster-id: app-name-"+hash+"\n"+
 			"taskmanager.host: $HOST_IP\n",
 			common.GetEnvVar(deployment.Spec.Template.Spec.Containers[0].Env,
-				"OPERATOR_FLINK_CONFIG").Value)
+				"FLINK_PROPERTIES").Value)
```
glaksh100: Shall we continue to test for `OPERATOR_FLINK_CONFIG` being populated here?

anandswaminathan:
Just to understand correctly, 

We are changing from `OPERATOR_FLINK_CONFIG ` to `FLINK_PROPERTIES`, and  keeping `OPERATOR_FLINK_CONFIG` for now for backward compatibility. Right ?

tweise:
@anandswaminathan yes, we can remove `OPERATOR_FLINK_CONFIG` after a couple of releases to allow users to migrate their Docker files.

pkg/controller/flink/task_manager_controller_test.go
tweise: Added the check.

glaksh100:
This change will cause applications to restart, right (since the hash has changed)?

tweise:
@glaksh100 thanks for pointing this out, a noted it in the commit message.

PR #151, opened 2020-01-03
[STRMCMP-437] Making the checkpoint restoration age configurable

premsantosh:
Adding a new option `maxCheckpointRestoreAgeSec` to configure checkpoint restoration age.

### DIFF
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -91,6 +91,9 @@ spec:
                   type: string
             savepointPath:
               type: string
+            maxCheckpointRestoreAgeSec:
+              type: integer
+              minimum: 1
             jobManagerConfig:
               type: object
               properties:
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -91,7 +91,7 @@ spec:
                   type: string
             savepointPath:
               type: string
-            maxCheckpointRestoreAgeSec:
+            maxCheckpointRestoreAgeSeconds:
               type: integer
               minimum: 1
             jobManagerConfig:
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -114,3 +114,5 @@ Below is the list of fields in the custom resource and their description
     Can be set to true to force rollback a deploy/update. The rollback is **not** performed when the application is in a **RUNNING** phase.
     If an application is successfully rolled back, it is moved to a *DeployFailed* phase. Un-setting or setting `ForceRollback` to `False` will allow updates to progress normally.
     
+  * **maxCheckpointRestoreAgeSeconds** `type:int32`
+    Used to prevent the application from restoring state from a checkpoint who's age (in seconds) is greater than the value set. It defaults to 1 hour (3600 seconds).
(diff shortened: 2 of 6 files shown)


pkg/apis/app/v1beta1/types.go
```diff
@@ -41,20 +42,21 @@ type FlinkApplicationSpec struct {
...
+	QueryPort             *int32                    `json:"queryPort,omitempty"`
+	UIPort                *int32                    `json:"uiPort,omitempty"`
+	MetricsQueryPort      *int32                    `json:"metricsQueryPort,omitempty"`
+	Volumes               []apiv1.Volume            `json:"volumes,omitempty"`
+	VolumeMounts          []apiv1.VolumeMount       `json:"volumeMounts,omitempty"`
+	RestartNonce          string                    `json:"restartNonce"`
+	DeleteMode            DeleteMode                `json:"deleteMode,omitempty"`
+	AllowNonRestoredState bool                      `json:"allowNonRestoredState,omitempty"`
+	ForceRollback         bool                      `json:"forceRollback"`
+	MaxCheckpointRestoreAgeSec time.Duration        `json:"maxCheckpointRestoreAgeSec,omitempty"`
```
mwylde: I'm not sure using the time.Duration type will do the right thing. You might need to make this an int32 and do the conversion manually.

deploy/crd.yaml
```diff
@@ -91,6 +91,9 @@ spec:
                   type: string
             savepointPath:
               type: string
+            maxCheckpointRestoreAgeSec:
```
mwylde: According to the [kubernetes api guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units), the unit in the duration field name should not be abbreviated (so `Seconds` instead of `Sec`).

pkg/controller/flink/flink.go
```diff
@@ -457,7 +454,7 @@ func (f *Controller) FindExternalizedCheckpoint(ctx context.Context, application
 		return "", nil
 	}
 
-	if time.Since(time.Unix(checkpoint.TriggerTimestamp, 0)) > maxRestoreCheckpointAge {
+	if time.Since(time.Unix(checkpoint.TriggerTimestamp, 0)) > (application.Spec.MaxCheckpointRestoreAgeSec * time.Second) {
```
mwylde: We will need a default value for when this is not set. See https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flink/config.go#L38 for an example of this.

premsantosh: Set the default to 1 hour.

docs/crd.md
```diff
@@ -114,3 +114,5 @@ Below is the list of fields in the custom resource and their description
     Can be set to true to force rollback a deploy/update. The rollback is **not** performed when the application is in a **RUNNING** phase.
     If an application is successfully rolled back, it is moved to a *DeployFailed* phase. Un-setting or setting `ForceRollback` to `False` will allow updates to progress normally.
     
+  * **maxCheckpointRestoreAgeSeconds** `type:int32`
+    Used to prevent the application from restoring state from a checkpoint who's age (in seconds) is greater than the value set. It defaults to 1 hour (3600 seconds).
```
mwylde: nit: whose not who's

Might also be good to add a bit of guidance about the implications of this config. I would also make clear that this is about using externalized checkpoints in the [operator update flow](https://github.com/lyft/flinkk8soperator/blob/master/docs/state_machine.md#savepointing), and doesn't affect flink's normal checkpoint recovery mechanism.

PR #153, opened 2020-01-07
[STRMCMP-771] Properly initialize http client to fix multiple submission issue

mwylde:
This fixes a bug that could lead to multiple job submissions in cases where there was an http-connection level error (e.g., a timeout).

Previously we were not calling `New()` on the resty package, which meant that we were using a single instance of the resty client across all calls. The meant that when we configured it for any call, the same configuration was used for subsequent calls. In particular, that meant that once we did a Get call, setting the number of retries to 3, that policy was applied to all subsequent calls, including POSTs, which should never be automatically retried. When a timeout occurred, this could then lead to multiple job submissions as the POST request was retried.


glaksh100:
Good find!

PR #162, opened 2020-01-15
[STRMCMP-752] Support recovering from externalized checkpoint even when JM is down

mwylde:
Currently, if the JobManager is down there is no good path to recovery. This PR fixes that by allowing deploys to continue even in this case (assuming that there is a valid externalized checkpoint that the operator has seen).

There are a few changes needed to make this work:
* We now start the Running phase by checking for an update, delaying the check on job status (which requires talking to the JM) until afterwards; this allows us to still transition even when the JM is down or unreachable
* We add a new field `lastCheckpointPath` to the JobStatus which stores the last successful checkpoint the operator has seen, so that we can find this even if the JM subsequently goes down
* We add a new phase `Recovering` to the state machine, which is responsible for attempting to recover using an externalized checkpoint. This isn't strictly necessary to implement the feature, but greatly simplifies the logic of the Savepointing phase.
* The FindExternalizedCheckpoint function will now use the latestCheckpointPath in the JobStatus if it is unable to query the externalized checkpoint from the JobManager


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -381,6 +349,40 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
...
+			fmt.Sprintf("Failed to recover with externalized checkpoint: %s", reason))
+		s.updateApplicationPhase(app, v1beta1.FlinkApplicationRollingBackJob)
+		return statusChanged, nil
+	}
+
+	// TODO: the old job might still be running at this point... we should maybe try to force cancel it
+	//       (but if the JM is unavailable, our options there might be limited)
+
+	// try to find an externalized checkpoint
+	path, err := s.flinkController.FindExternalizedCheckpoint(ctx, app, app.Status.DeployHash)
```
anandswaminathan: Should we have this behind a config or setting ?

docs/state_machine.mmd
```diff
@@ -18,7 +18,10 @@ ClusterStarting --> Savepointing
 ClusterStarting -- Create fails --> DeployFailed
 
 Savepointing --> SubmittingJob
-Savepointing -- Savepoint fails and no externalized checkpoint --> DeployFailed
+Savepointing -- Savepoint fails --> Recovering
+
+Recovering --> SubmittingJob
```
anandswaminathan: Nit:

You might have put more thought into this ? `Recovering` is very generic, and explains the step. But since this is tied to Savepoint now, do you want something specific?  Or are you planning to use this state for other use cases.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -381,6 +349,40 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
 	return statusUnchanged, nil
 }
 
+func (s *FlinkStateMachine) handleApplicationRecovering(ctx context.Context, app *v1beta1.FlinkApplication) (bool, error) {
+	// we're in the middle of a deploy, and savepointing has failed in some way... we're going to try to recover
+	// and push through if possible
+	if rollback, reason := s.shouldRollback(ctx, app); rollback {
+		// we failed to recover, attempt to rollback
+		s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobsSubmissionFailed",
```
glaksh100: Should this event be something other than `JobsSubmissionFailed`?

pkg/controller/flinkapplication/flink_state_machine.go
mwylde: This isn't changing anything from the current behavior, just moving things around.

docs/state_machine.mmd
mwylde: "Recover" is what we call this currently in our legacy platform. Happy to hear other suggestions though.

pkg/controller/flinkapplication/flink_state_machine.go
mwylde: Yes, good call; updated to "RecoveryFailed"

PR #171, opened 2020-02-19
[STRMCMP-825] Improve error messaging when moving to DeployFailed in a non-recoverable state

mwylde:
Currently, all transitions to deployFailed production a `Successfully rolled back deploy` event (via the FlinkStateMachine.deployFailed method). However, there are some cases where the rollback in fact has failed; for example in the sequence where
1. cancelling the existing job succeeds, but we fail to find the savepoint path for some reason
2. there's never been a successful checkpoint taken for the job (or we fail to query it), so recovery fails

This change improves the error message in those cases to be clearer to the user that rollback has failed, and manual intervention is needed to recover.

Also includes a nil-pointer deference bug fix for the output of `GetJobForApplication`, which I encountered while reproducing this.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -298,7 +298,6 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
 		s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "SavepointFailed",
 			fmt.Sprintf("Could not savepoint existing job: %s", reason))
 		application.Status.RetryCount = 0
-		application.Status.JobStatus.JobID = ""
 		s.updateApplicationPhase(application, v1beta1.FlinkApplicationRecovering)
 		return statusChanged, nil
 	}
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -231,7 +231,7 @@ func (s *FlinkStateMachine) handleNewOrUpdating(ctx context.Context, application
 		// we've failed to make progress; move to deploy failed
 		s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "ClusterCreationFailed",
 			fmt.Sprintf("Failed to create Flink Cluster: %s", reason))
-		return s.deployFailed(ctx, application)
+		return s.deployFailed(application)
 	}
 
 	// Create the Flink cluster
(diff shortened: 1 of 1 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -301,6 +298,7 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
 		s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "SavepointFailed",
 			fmt.Sprintf("Could not savepoint existing job: %s", reason))
 		application.Status.RetryCount = 0
+		application.Status.JobStatus.JobID = ""
```
anandswaminathan: Any reason to clear the job id, when savepointing is not succeeding?

mwylde: Whoops, that change should not have been left in. Good catch!

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -512,6 +514,9 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
 	if err != nil {
 		return statusUnchanged, err
 	}
+	if job == nil {
+		return statusUnchanged, errors.Errorf("Could not find job %s", app.Status.JobStatus.JobID)
```
glaksh100: Not directly related to the change in the PR, but I'm curious how we end up here. It looks like `GetJobForApplication` only returns a `nil` job when the jobID is empty. If jobID were empty we should end up in the `if app.Status.JobStatus.JobID == "" ` above and not here? Basically curious, how you ran into the nil deference issue and if there's another bug that's being masked. We can discuss outside of the PR too :)

mwylde: `flinkClient.GetJobOverview` has this error handling logic:

```go
var jobOverviewResponse FlinkJobOverview
if err = json.Unmarshal(response.Body(), &jobOverviewResponse); err != nil {
	logger.Errorf(ctx, "Failed to unmarshal FlinkJob %v, err %v", response, err)
}

return &jobOverviewResponse, nil
```

Which means that if we fail to unmarshal the response, we end up getting `nil, nil`.

mwylde: Actually, that explanation is wrong—we'll get an empty FlinkJobOverview struct, not nil. I think your'e right that the check should not be needed here.

glaksh100: Ah okay, that makes sense. Thanks for the context. :) Yeah, maybe `GetJobOverview`'s error response should be changed..

PR #172, opened 2020-02-19
[STRMCMP-844] Fix rollback when SubmittingJob times out

mwylde:
If we time out after the new job has been submitted (generally because the tasks are never all running, possibly due to parallelism being misconfigured), we will enter the rollback phase. However, because the new job _was_ submitted, the jobID is set in our status, and so [this check|https://github.com/lyft/flinkk8soperator/blob/4142437353666b8692e62acf075a9b2c70514dd9/pkg/controller/flinkapplication/flink_state_machine.go#L396] prevents us from submitting the job to the old cluster.

This PR fixes that issue by clearing the jobId before moving to the RollingBack phase.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -460,6 +460,7 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
 		// Something's gone wrong; roll back
 		s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobSubmissionFailed",
 			fmt.Sprintf("Failed to submit job: %s", reason))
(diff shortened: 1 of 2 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -460,6 +460,7 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
 		// Something's gone wrong; roll back
 		s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobSubmissionFailed",
 			fmt.Sprintf("Failed to submit job: %s", reason))
+		app.Status.JobStatus.JobID = ""
```
anandswaminathan: Want to add a small unit test?

mwylde:
Added a test

PR #184, opened 2020-03-06
Allow disabling Savepoint during updates

maghamravi:
The PR 

1. Adds support for a new field `SavepointDisabled` to the spec to govern if Savepoints are needed during an update.
2. Renames the state / phase `FlinkApplicationSavepointing` to `FlinkApplicationCancelling`.
3. Recovery workflow (if and when Cancel fails) remain the same. 
3. No changes to the Delete workflow. Users will have to set `deleteMode: ForceCancel`  if they prefer not having Savepoints taken during Delete/cancel.


maghamravi:
/ptal (👀) @mwylde @glaksh100 @anandswaminathan @kumare3 @tweise

docs/state_machine.mmd
```diff
@@ -14,11 +14,11 @@ Running --> Updating
 Updating --> ClusterStarting
 DeployFailed --> Updating
 
-ClusterStarting --> Savepointing
+ClusterStarting --> Cancelling
 ClusterStarting -- Create fails --> DeployFailed
 
-Savepointing --> SubmittingJob
-Savepointing -- Savepoint fails --> Recovering
+Cancelling --> SubmittingJob
```
anandswaminathan: @maghamravi Not a fan of this approach.

Ideally I would not change the existing statemachine as much as possible. If just cancel is needed without savepointing, my recommendation would be to introduce a new state, say "Cancelling" that is reached from Savepointing if savepoint is disabled in spec. I believe "Recovering" is also not associated with Cancelling.

maghamravi: Honestly, I think we clubbed "Cancel" and "Savepoint" (two independent actions) into one called "Savepointing" which IMO should be "Cancelling". 
  
Dunno if it's just me(I think?),  but  when I see the state "Savepointing', I am under the impression that the operator is triggering a "Savepoint" on the job (https://ci.apache.org/projects/flink/flink-docs-stable/monitoring/rest_api.html#jobs-jobid-savepoints ) and not Cancelling it (with the **cancels** option)
 
If the consensus is to have "Cancel" be called from Savepointing in cases when "savepoint" is disabled,  though I disagree,  I am happy to commit!

anandswaminathan: Agreed. There are two things here:

1) Better name for the Savepointing state.
2) Separating the intention of this PR to a separate phase/state.

I merely am recommending (2) now. We can always revisit (1) and update the state names later.

maghamravi:
Updated the PR with the following

1. Created a new state "Cancelling".
2. From "ClusterStarting" state, based on "SavepointDisabled" field in the spec, the state machine transitions to  "Cancelling" or "Savepointing" state.
3.  Though cancelling a job (without save point) rarely fails, I have it go through the max retries. If we are still unlucky,  we move to "Submitting" phase as by then current running job(if existing) is already dysfunctional.

docs/state_machine.mmd
maghamravi: @anandswaminathan Made the necessary changes!

docs/state_machine.mmd
```diff
@@ -14,9 +14,11 @@ Running --> Updating
 Updating --> ClusterStarting
 DeployFailed --> Updating
 
-ClusterStarting --> Savepointing
+ClusterStarting -- savepoint disabled --> Cancelling
```
anandswaminathan: Remove "savepoint disabled" and enabled as these not states.

anandswaminathan:
cc @tweise 

@glaksh100 This change is small but you might need to add a check for your Green-Blue update.

@maghamravi This looks good.
Add to 
1) https://github.com/lyft/flinkk8soperator/blob/master/docs/crd.md
2) Validation here: https://github.com/lyft/flinkk8soperator/blob/master/deploy/crd.yaml#L23
3) Integration test for this change - as you will be the first using this.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -346,6 +352,34 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
...
+		return statusChanged, nil
+	}
+
+	if rollback, reason := s.shouldRollback(ctx, application); rollback {
+		// it's rare to come here despite the retries. let's head to submitting.
+		s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "CancelFailed",
+			fmt.Sprintf("Could not cancel existing job: %s", reason))
+		application.Status.RetryCount = 0
+		application.Status.JobStatus.JobID = ""
+		s.updateApplicationPhase(application, v1beta1.FlinkApplicationSubmittingJob)
```
anandswaminathan: This should be `FlinkApplicationRollingBackJob`

We can't move ahead and submit job in the second cluster, if the first cluster is still running.

pkg/controller/flinkapplication/flink_state_machine_test.go
```diff
@@ -81,6 +81,132 @@ func TestHandleStartingClusterStarting(t *testing.T) {
 	assert.Nil(t, err)
 }
 
+func TestHandleNewOrCreateWithSavepointDisabled(t *testing.T) {
```
mwylde: This is great, but can you also add an integration test? Given this is a completely different flow, would be good to fully test out the successful and failure paths.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -346,6 +352,34 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
...
+		return statusChanged, nil
+	}
+
+	err := s.flinkController.ForceCancel(ctx, application, application.Status.DeployHash)
```
mwylde: I think there's a potential case here where we get an error back from this call (e.g., a timeout), but the job does end up cancelled. We can handle that by first checking whether the job is running, and if not, moving on to SubmittingJob.

pkg/apis/app/v1beta1/types.go
```diff
@@ -220,6 +221,7 @@ const (
 	FlinkApplicationSubmittingJob   FlinkApplicationPhase = "SubmittingJob"
 	FlinkApplicationRunning         FlinkApplicationPhase = "Running"
 	FlinkApplicationSavepointing    FlinkApplicationPhase = "Savepointing"
+	FlinkApplicationCancelling      FlinkApplicationPhase = "Cancelling"
```
mwylde: Can you also update state_machine.md with the details of this new state?

pkg/controller/flinkapplication/flink_state_machine.go
maghamravi: In my testing, I saw two kinds of errors

1. status code 404 when the job is already in Cancelled status
2. status code 400 for a bad request like the job with Id doesn't exist. 

It definitely doesn't hurt to make that extra call to check if the job is running or not.

pkg/controller/flinkapplication/flink_state_machine.go
maghamravi: Good point! Definitely better to err on the side of caution.

pkg/controller/flinkapplication/flink_state_machine.go
mwylde: Yeah, in general for operators the pattern is (1) query the current state of the world; (2) make calls to update the world to the desired state. On each iteration of the reconciliation loop you don't really know what the state is until you query it (even the status might out of date or missing updates).

pkg/apis/app/v1beta1/types.go
maghamravi: done!

pkg/controller/flinkapplication/flink_state_machine_test.go
maghamravi: done!

pkg/controller/flinkapplication/flink_state_machine.go
maghamravi: done!

maghamravi:
@anandswaminathan @mwylde Addressed all review comments!

anandswaminathan:
+1 LGTM
cc @mwylde @glaksh100

docs/crd.md
```diff
@@ -90,6 +90,9 @@ Below is the list of fields in the custom resource and their description
   * **savepointPath** `type:string`
     If specified, the application state will be restored from this savepoint
 
+  * **savepointDisabled** `type:boolean`
+    If specified, the current application (if existing) is cancelled without taking a savepoint.
```
mwylde: Might be good to make it clear that this applies only during updates

maghamravi: done!

PR #194, opened 2020-04-22
Fix status update post-teardown

glaksh100:
This PR fixes a bug in that manifests in the following way:
- BlueGreen deploy happens and application is in a DualRunning phase
- TearDownVersionHash is set and the version corresponding to the hash is torn down and application is in a Running phase
- Subsequent update happens and application moves back to DualRunning phase. However, since there's an obsolete tearDownVersionHash set. While this keeps the application in a DualRunning state by ignoring the obsolete hash, it does not execute the entire block of `handleDualRunning()` and instead returns here.https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L948 
- This implies that the cluster/job statuses are no longer updated: https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L952

This is fixed by moving the check for an obsolete hash into the `handleDualRunning()`method, logging a warning if there's no matching version to tear down, but otherwise continuing with the job/cluster status updates.


glaksh100:
PTAL @mwylde @anandswaminathan  when you have a chance!

PR #199, opened 2020-05-04
Update with external checkpoints

davidbirdsong:
Adding in values to the CRD to support conveying that savepoint during upgrade can be performed with a checkpoint instead.

Checkpoints are running constantly in the background and so a 'recent enough' checkpoint must exist and that recency configurable.


pkg/apis/app/v1beta1/types.go
```diff
@@ -58,6 +59,7 @@ type FlinkApplicationSpec struct {
 	AllowNonRestoredState          bool                `json:"allowNonRestoredState,omitempty"`
 	ForceRollback                  bool                `json:"forceRollback"`
 	MaxCheckpointRestoreAgeSeconds *int32              `json:"maxCheckpointRestoreAgeSeconds,omitempty"`
+	MaxCheckpointDeployAgeSeconds  *int32              `json:"maxCheckpointDeployAgeSeconds,omitempty"`
```
glaksh100, line 62: Is `MaxCheckpointDeployAgeSeconds` functionally different from the `MaxCheckpointRestoreAgeSeconds`?

glaksh100:
Hi @davidbirdsong ! Thank you for creating this PR. 

I have one overall comment/suggestion on the approach. Including the functionality you have in this PR, we have 3 ways to perform an update today: WithSavepoint, Without Savepoint, With Checkpoint. I'm wondering if we can consolidate these options into one field (maybe something called `UpdateMode` etc)?. 

We will still have to support the existing flags for backward compatibility. Thoughts?
cc. @mwylde

pkg/apis/app/v1beta1/types.go
davidbirdsong: Not functionally, no. my thinking was that these checkpoint age values could have very different acceptability thresholds since the `MaxCheckpointDeployAgeSeconds` would be looking for an ideal, possibly short window. I may have read the code wrong, but my thinking was by failing this age check, the operator would effectively poll the JM API looking for the next checkpoint to run and complete (we have ours checkpointing frequently.)

I may have read the state machine handler incorrectly though.

davidbirdsong:
> Hi @davidbirdsong ! Thank you for creating this PR.
> 
> I have one overall comment/suggestion on the approach. Including the functionality you have in this PR, we have 3 ways to perform an update today: WithSavepoint, Without Savepoint, With Checkpoint. I'm wondering if we can consolidate these options into one field (maybe something called `UpdateMode` etc)?.
> 
> We will still have to support the existing flags for backward compatibility. Thoughts?
> cc. @mwylde

I figured additive was much easier WRT to backwards compatibility since savepoints are part of job cancellation while checkpoints are not and so the supporting behavior savepoints and checkpoints would be very different.

I'm open to unifying the functionality/implementation.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -315,6 +315,18 @@ func (s *FlinkStateMachine) handleClusterStarting(ctx context.Context, applicati
 	}
 	return statusChanged, nil
 }
+func (s *FlinkStateMachine) handleApplicationSavepointingWithCheckpoint(ctx context.Context, application *v1beta1.FlinkApplication) (bool, error) {
+	checkpointPath, err := s.flinkController.FindExternalizedCheckpointForSavepoint(ctx, application, application.Status.DeployHash)
```
glaksh100, line 319: Cancelling the job would still be required here, correct?

davidbirdsong: We still need to cancel, yes, and maybe I didn't read closely enough but: `handleApplicationSavepointingWithCheckpoint` is called within `handleApplicationSavepointing` which also does not cancel. I figured this area of code was only responsible for ensuring a valid savepoint path had been harvested and stored on `application.Status.SavepointPath`.

I may be wrong in invoking the new function in `handleApplicationSavepointing` since savepoints can be requested directly on the cancel API call.

I'll review again.

pkg/controller/flink/client/api.go
```diff
@@ -66,49 +67,53 @@ type FlinkJobManagerClient struct {
...
+		getClusterSuccessCounter:           labeled.NewCounter("get_cluster_success", "Get cluster overview succeeded", flinkJmClientScope),
+		getClusterFailureCounter:           labeled.NewCounter("get_cluster_failure", "Get cluster overview failed", flinkJmClientScope),
+		getCheckpointsSuccessCounter:       labeled.NewCounter("get_checkpoints_success", "Get checkpoint request succeeded", flinkJmClientScope),
+		getCheckpointsFailureCounter:       labeled.NewCounter("get_checkpoints_failed", "Get checkpoint request failed", flinkJmClientScope),
+		savepointJobSuccessCounter:         labeled.NewCounter("savepoint_job_success", "Savepoint job request succeeded", flinkJmClientScope),
+		savepointJobFailureCounter:         labeled.NewCounter("savepoint_job_failed", "Savepoint job request failed", flinkJmClientScope),
+		getCheckpointsConfigSuccessCounter: labeled.NewCounter("get_checkpoints_config_success", "Get checkpoint config request succeeded", flinkJmClientScope),
+		getCheckpointsConfigFailureCounter: labeled.NewCounter("get_checkpoints_config_failed", "Get checkpoint config request failed", flinkJmClientScope),
 	}
```
davidbirdsong, line 117: I love `go fmt`, but it can also be disruptive causing PR's to bloat changes and hides the actual diff. 

I could add some line breaks to separate out the new, longer struct member and minimize the line change count on things like this.

glaksh100:
Sorry for the delay here @davidbirdsong ! Thanks for addressing the review comments. I am leaning on the approach where we introduce an `UpdateMode`, which can default to the savepoint and explicitly use either of the other two approaches (checkpoint or `savepointDisabled`). The only flag we'd need to retain for backward compatibility would be `savepointDisabled`, which IMO is okay. Let me know what you think!

cc/ @anandswaminathan @mwylde

davidbirdsong:
> Sorry for the delay here @davidbirdsong ! Thanks for addressing the review comments. I am leaning on the approach where we introduce an `UpdateMode`, which can default to the savepoint and explicitly use either of the other two approaches (checkpoint or `savepointDisabled`). The only flag we'd need to retain for backward compatibility would be `savepointDisabled`, which IMO is okay. Let me know what you think!
> 
> cc/ @anandswaminathan @mwylde

Can you advise on how to handle the coupling of needing both of the following to be true:
- externalized checkpoints enabled
- `RETAIN_ON_CANCELLATION` set on the job env itself


without these, any checkpoint-based update will fail. For this PR, I kept it simple by returning from `handleApplicationSavepointingWithCheckpoint` with the state machine unchanged, but I'm not clear on what good behavior should be. The requirement for the setting and the setting itself are far apart: operator vs app and could lead to a deadlock of sorts. What do you recommend?

glaksh100:
Hey @davidbirdsong ! Thanks for the updates. I will review this PR in the next couple of days :)

pkg/controller/flink/flink.go
```diff
@@ -490,12 +493,28 @@ func (f *Controller) FindExternalizedCheckpoint(ctx context.Context, application
...
 	return checkpointPath, nil
+
+}
+
+func (f *Controller) FindExternalizedCheckpoint(ctx context.Context, application *v1beta1.FlinkApplication, hash string) (string, error) {
+	return f.findExternalizedCheckpoint(ctx, application, hash, getMaxCheckpointRestoreAgeSeconds(application))
+}
+
+func (f *Controller) FindExternalizedCheckpointForSavepoint(ctx context.Context, application *v1beta1.FlinkApplication, hash string) (string, error) {
+	checkpointConfig, err := f.flinkClient.GetCheckpointConfig(ctx, f.getURLFromApp(application, hash), application.Status.JobStatus.JobID)
```
glaksh100, line 510: Is there opportunity to combine `FindExternalizedCheckpoint` and `FindExternalizedCheckpointForSavepoint` into a single method? I believe that the checkpoint configuration check (to ensure it's RETAIN_ON_CANCELLATION) is applicable to both cases.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -342,6 +362,11 @@ func (s *FlinkStateMachine) handleApplicationSavepointing(ctx context.Context, a
 		s.updateApplicationPhase(application, v1beta1.FlinkApplicationRecovering)
 		return statusChanged, nil
 	}
+	// use of checkpoints in the place of savepoints
+	if application.Spec.UpdateMode == v1beta1.UpdateModeCheckpoint {
+		return s.handleApplicationSavepointingWithCheckpoint(ctx, application)
```
glaksh100, line 367: Do you think it would make sense to make the restoration from a checkpoint an explicit state in the state machine?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -305,16 +306,34 @@ func (s *FlinkStateMachine) handleClusterStarting(ctx context.Context, applicati
...
-	} else if application.Spec.SavepointDisabled && v1beta1.IsBlueGreenDeploymentMode(application.Status.DeploymentMode) {
+	} else if application.Spec.SavepointingDisabled() && v1beta1.IsBlueGreenDeploymentMode(application.Status.DeploymentMode) {
 		// Blue Green deployment and no savepoint required implies, we directly transition to submitting job
 		s.updateApplicationPhase(application, v1beta1.FlinkApplicationSubmittingJob)
 	} else {
 		s.updateApplicationPhase(application, v1beta1.FlinkApplicationSavepointing)
 	}
 	return statusChanged, nil
 }
+func (s *FlinkStateMachine) handleApplicationSavepointingWithCheckpoint(ctx context.Context, application *v1beta1.FlinkApplication) (bool, error) {
```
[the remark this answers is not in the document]
glaksh100: Also minor suggestion: Could we rename this method to not include the term `Savepoint`(since it's not really using a savepoint)?

pkg/controller/flinkapplication/flink_state_machine.go
davidbirdsong: Hmm, my gut was that it doesn't warrant it, but I suppose that since there's a max-age and a window where the controller is polling to find the next, fresher checkpoint that the controller might do well to expose that state and have a clearer exit condition. I can prototype it, if you'd like.

glaksh100: Yeah I agree that it would be clearer to expose it as a state. Happy to help review the prototype!

glaksh100:
That's a very good point! I don't believe we'd want the `updateMode` and `savepointPath` to interact at all in fact, because the expectation today is that the user-provided `savepointPath` is only used during the first deploy of the application (see: https://github.com/lyft/flinkk8soperator/blob/master/pkg/controller/flinkapplication/flink_state_machine.go#L578)

In the first deploy of a job, we use the provided savepoint (if any) while submitting the job. During an update, however the field is not used. So I think we should be okay with keeping these states/transitions isolated. Let me know what you think.

Another thought that came to mind is that, since v0.5 we have a `BlueGreen` dpeloyment mode that allows a new job to be submitted (without cancelling the old job). During the update in the mode, we take a savepoint of the old job (w/o cancellation). We should ensure that the `BlueGreen` deployment mode behavior is preserved in all updateModes.

PR #205, opened 2020-06-19
Add service options

vDMG:
Allow users to specify serviceType (essential if you are working with Google Kubernetes Engine)
Allow users to specify serviceAnnotations (essential if you are working with Google Kubernetes Engine)

### DIFF
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -40,6 +40,12 @@ spec:
                 properties:
                   name:
                     type: string
+            serviceAnnotations:
+              description: May be used to configure the service objects
+              type: object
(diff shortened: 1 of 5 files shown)


vDMG:
@kumare3 can you take a look plz :) ?

anandswaminathan:
@vDMG 

Hi Vince, We have tried to not include non flink fields as much as possible from the CRD spec, but in cases have not been able to avoid it.

This change is very specific to the JobManager service that get's created. It messes up more if we change up the way service gets setup. The recommendation we provide is to just employ a simple mutation webhook on your end to achieve this. Did you try that approach ?

PR #207, opened 2020-07-17
Adapt to Flink v1.11 memory configuration

alberto-quesada-scopely:
The memory setup has changed a lot with the 1.10 release for TaskManagers and with the 1.11 release for JobManagers.

The options set by the operator (`jobmanager.heap.size` and `taskmanager.heap.size`) are now deprecated. They can still be used in order to maintain backwards compatibility but they will be interpreted as a different option. Full migration guide can be found [here](https://ci.apache.org/projects/flink/flink-docs-master/ops/memory/mem_migration.html#migration-guide).

With this change, when the version of Flink that we are deploying is lower than 1.11 nothing will change: memory options will be computed as it was before.
When the Flink version is 1.11 or above, we will set `jobmanager.memory.process.size` and `taskmanager.memory.process.size` using the user defined requested memory.

If no Flink version is specified or we set an incorrect version (non semantic versioning model for example), default configuration will be set.

A new dependency has been added: [go-version](https://github.com/hashicorp/go-version). It helps parsing semantic versioning and comparing between versions.

### DIFF
--- a/deploy/crd.yaml
+++ b/deploy/crd.yaml
@@ -141,6 +141,10 @@ spec:
                   type: number
                   minimum: 0
                   maximum: 1
+                systemMemoryFraction:
+                  type: number
+                  minimum: 0
+                  maximum: 1
                 nodeSelector:
                   type: object
                   properties:
@@ -281,6 +285,10 @@ spec:
                   type: integer
                   minimum: 1
                 offHeapMemoryFraction:
+                  type: number
+                  minimum: 0
+                  maximum: 1
+                systemMemoryFraction:
                   type: number
                   minimum: 0
                   maximum: 1
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -39,8 +39,7 @@ Below is the list of fields in the custom resource and their description:
       Number of task slots per task manager.
 
     * **systemMemoryFraction** `type:float64`
-      A value between 0 and 1 that represents % of container memory dedicated to system / off heap. The
-      remaining memory is given to the taskmanager.
+      A value between 0 and 1 that represents % of container memory dedicated to the system. The remaining memory is given to the taskmanager.
 
     * **nodeSelector** `type:map[string]string`
       Configuration for the node selectors used for the task manager.
@@ -63,8 +62,7 @@ Below is the list of fields in the custom resource and their description:
       correct environment variables are set for High availability mode.
 
     * **systemMemoryFraction** `type:float64`
-      A value between 0 and 1 that represents % of container memory dedicated to system / off heap. The
-      remaining memory is given to the job manager.
+      A value between 0 and 1 that represents % of container memory dedicated to the system. The remaining memory is given to the job manager.
 
     * **nodeSelector** `type:map[string]string`
       Configuration for the node selectors used for the job manager.
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -38,11 +38,9 @@ Below is the list of fields in the custom resource and their description:
     * **taskSlots** `type:int32 required=true`
       Number of task slots per task manager.
 
-    * **offHeapMemoryFraction** `type:float64`
+    * **systemMemoryFraction** `type:float64`
       A value between 0 and 1 that represents % of container memory dedicated to system / off heap. The
-      remaining memory is given to the taskmanager. Note that Flink may further reserve some of this
-      memory for off-heap uses like network buffers, so you may see the JVM heap size configured to
-      a lower amount.
+      remaining memory is given to the taskmanager.
 
     * **nodeSelector** `type:map[string]string`
       Configuration for the node selectors used for the task manager.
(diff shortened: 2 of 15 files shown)


kelly-sm:
Shouldn't this logic be applied for TM in version >=1.10? What's the reasoning behind only adding it for >= 1.11?

alberto-quesada-scopely:
> Shouldn't this logic be applied for TM in version >=1.10? What's the reasoning behind only adding it for >= 1.11?
> 
> Without this logic in place in 1.10 - I have to manually configure `taskmanager.memory.process.size` to match with the computed heap size or else I get validation errors.

Only half of this configuration is valid for the version 1.10. In the version 1.11 they added the same behaviour to the JM for consistency. 
We could change this so when version is >= 1.10, we would set `taskmanager.memory.process.size` and when version is >= 1.11 we add `jobmanager.memory.process.size` as well.

kelly-sm:
> > Shouldn't this logic be applied for TM in version >=1.10? What's the reasoning behind only adding it for >= 1.11?
> > Without this logic in place in 1.10 - I have to manually configure `taskmanager.memory.process.size` to match with the computed heap size or else I get validation errors.
> 
> Only half of this configuration is valid for the version 1.10. In the version 1.11 they added the same behaviour to the JM for consistency.
> We could change this so when version is >= 1.10, we would set `taskmanager.memory.process.size` and when version is >= 1.11 we add `jobmanager.memory.process.size` as well.

That makes more sense to me, just wanted to confirm whether I was missing something/misread the Flink docs.

mwylde:
Hey @alberto-quesada-scopely — have you had a chance to look at my comment? At Lyft we're also looking to roll out Flink 1.11 so would be great to get this change in.

alberto-quesada-scopely:
> Hey @alberto-quesada-scopely — have you had a chance to look at my comment? At Lyft we're also looking to roll out Flink 1.11 so would be great to get this change in.

Hey! I've been on holidays but I will try to adapt it asap ;)

alberto-quesada-scopely:
Hi @mwylde I just pushed some changes, but I'm not super happy with the name `offProcessMemoryFraction` either... is hard to come up with something that works fine for both off heap and off process...
Other than that, I also added support for process memory in task managers for version 1.10.

pkg/controller/flink/config.go
```diff
@@ -124,8 +129,21 @@ func renderFlinkConfig(app *v1beta1.FlinkApplication) (string, error) {
...
 	(*config)["metrics.internal.query-service.port"] = getInternalMetricsQueryPort(app)
-	(*config)["jobmanager.heap.size"] = getJobManagerHeapMemory(app)
-	(*config)["taskmanager.heap.size"] = getTaskManagerHeapMemory(app)
+
+	appVersion, err := version.NewVersion(getFlinkVersion(app))
+	v10, _ := version.NewVersion("1.10")
+	v11, _ := version.NewVersion("1.11")
+
+	if err != nil || appVersion == nil || appVersion.LessThan(v10) {
+		(*config)["jobmanager.heap.size"] = getJobManagerMemory(app)
```
mwylde, line 138: I think we need to be careful about conflating these configs (heap.size and process.size), since they control different things and will lead to different behavior when upgrading.

For existing users (who are on Flink <= 1.10 and using offHeapMemoryFraction), the behavior should stay the same; i.e., we shouldn't move to configuring the process size—we should only do that if they specify the new config.

deploy/crd.yaml
```diff
@@ -134,7 +134,7 @@ spec:
                 replicas:
                   type: integer
                   minimum: 1
-                offHeapMemoryFraction:
+                offProcessMemoryFraction:
```
mwylde, line 137: For backwards compatibility, we'll need to also keep the existing "offHeapMemoryFraction" field in the CRD definition and throughout the codebase. Otherwise, this change will break existing users who are relying on that. We can mark it as deprecated in the docs, and if both are specified we can ignore offHeap. See what we've done for 

 We can mark it deprecated in the docs though.

Naming-wise, maybe "nonFlinkMemoryFraction"? Or we could invert it, and have it as something like "systemMemoryFraction."

pkg/controller/flink/config.go
alberto-quesada-scopely: Ok I kept the same configuration for versions <= 1.10, thanks!

deploy/crd.yaml
alberto-quesada-scopely: Can you take a look at the new changes? I think it should be ok now :)

docs/crd.md
```diff
@@ -38,11 +38,9 @@ Below is the list of fields in the custom resource and their description
     * **taskSlots** `type:int32 required=true`
       Number of task slots per task manager
 
-    * **offHeapMemoryFraction** `type:float64`
+    * **systemMemoryFraction** `type:float64`
       A value between 0 and 1 that represents % of container memory dedicated to system / off heap. The
```
mwylde, line 42: I think with the new memory configs (`process.size`) off-heap is actually within the flink portion, rather than part of the system portion (I'm following the diagram [here](https://ci.apache.org/projects/flink/flink-docs-release-1.11/ops/memory/mem_setup.html#configure-total-memory)), so this should just read "% of container memory dedicated to the system."

pkg/controller/flink/config.go
```diff
@@ -17,7 +18,7 @@ const (
 	BlobDefaultPort                = 6125
 	UIDefaultPort                  = 8081
 	MetricsQueryDefaultPort        = 50101
-	OffHeapMemoryDefaultFraction   = 0.5
+	SystemMemoryDefaultFraction    = 0.5
```
mwylde, line 21: I think this should be lowered, since the new config includes off-heap. Maybe set this to 0.2 by default (while keeping the existing default for off-heap)?

mwylde:
Looks like there's a merge conflict — can you rebase on top of master?

alberto-quesada-scopely:
> Looks like there's a merge conflict — can you rebase on top of master?

Done! Thanks!

PR #209, opened 2020-07-21
Add InPlace scale mode

mwylde:
This PR gives the operator the ability to scale up a flink job "in place," which has the potential to be much faster than our current new cluster approach, helping to quickly mitigate production issues caused by under-scale. This feature is off by default, and can be enabled by setting the new `RescaleMode` option in the FlinkApplication spec to `InPlace`. This option should be considered experimental and is not recommended yet for production.

## Background

Currently, a change in parallelism (say from P → P’) is treated the same as any other change to the CRD, triggering a full deploy, with the following steps:

1. Create a new Flink cluster with size P’
2. Take a savepoint of the existing job
3. Cancel the existing job
4. Submit the new job to the new cluster
4. Tear down the old cluster

On our clusters, this takes an average of 5 minutes but can take as long as 15-20 minutes if there are not enough existing Kubernetes resources to fit the new cluster, requiring us to spin up new EC2 nodes.

However, for rescale operations (that is, a configuration change that only updates the parallelism field of the CRD) we can make this faster by reusing the existing cluster.

If P’ > P (i.e., this is a scale up), the new procedure looks like:
1. Update the size of the taskmanager deployment to P’; wait until all nodes are up and registered in the jobmanager
2. Cancel the job with savepoint
3. Resubmit the job to the cluster with the new size

The P’ < P (scale down) case presents more challenges, because we cannot remove resources while a job is running. If we scale down while the old job is running, we would be unable to savepoint it. If we scale down while the new job is running, we will incur a period of downtime as nodes go down and tasks are constantly being moved to other nodes (some of which may themselves be terminated). So an optimized downscale procedure would look like

1. Cancel job with savepoint
2. Update the size of the taskmanager deployment to P’; wait until all terminated nodes are gone from the jobmanager
3. Submit the job with the new parallelism

Between steps 1 and 2, there will be downtime, likely exceeding that of a normal deploy. Even worse, if the resubmit fails we cannot immediately rollback to the previous config because the nodes have already been torn down.

Because of these issues, and the generally lower time sensitivity of scale downs, we will only perform this optimization in the scale up case. We will also not do this in the case of blue/green deploys, because it violates the no-downtime property of blue/green.

## Implementation

The operator currently is architected around a core assumption that every deploy will create a new cluster. This assumption is encoded in the underlying resources by our heavy use the application hash (a hash made up of most fields of the application spec, used to determine whether something has changed). For example, we determine what service to use for a particular version of the flink application, the current set of deployments, etc.

But to re-use the same cluster (the existing pods and deployments) for an application whose spec has changed (its parallelism in this case) we have to find a way to "upgrade" those resources to the new hash. Pods are immutable, and the ultimate goal is to retain the existing pods, so we must leave them with their existing configuration. For deployments, we update the flink-app-hash (which is now an annotation instead of a label, as labels can't be changed) to transmute the deployments into our new set. We create a new service with the new hash that will be used by newly created pods.

But this presents an issue for our services, both old and new must continue working during the whole upgrade process. Currently we assign pods to the versioned service by the flink-app-hash label, which will not be consistent across old/new pods. So instead we now use a different, random value to connect services and pods, `pod-deployment-selector`. When we create the new service, we re-use this value so that the new service can address old and new pods.

## Compatibility

This change is compatible with the existing operator, which is to say that when users upgrade to this they will not see any changes to their applications. Updates and new applications will have slightly different Deployment configuration, but there should be no functional change unless users set the RescaleMode to InPlace.

## Benchmarks

There are a bunch of factors that will determine how much faster an in-place scale up is compared to NewCluster (the default). When scaling up from N -> N' pods, InPlace will require only `N' - N` pods be created, while NewCluster will require `N'` new pods. The main benefit comes when there is enough cluster capacity for `M - N` pods, but would not be for `N`. However, we have also seen empirically on our clusters that it takes longer to spin up more nodes, so requiring fewer should be faster in general as well.

I ran some experiments on a Lyft cluster, rescaling a simple application from 8 -> 12 pods.

In the case where there is not enough capacity for 12 new pods, but is for 4, I found average timings for scale-up as follows:

| NewCluster | InPlace | Speed up |
| ------------|---------|----------|
| 5:59       | 1:41    | 72%      |

The actual impact of this change however will likely depend heavily on the cluster setup.


pkg/apis/app/v1beta1/types.go
```diff
@@ -183,6 +184,7 @@ type FlinkApplicationStatus struct {
 	DeployHash         string                          `json:"deployHash"`
 	UpdatingHash       string                          `json:"updatingHash,omitempty"`
 	TeardownHash       string                          `json:"teardownHash,omitempty"`
+	InPlaceUpdatedFrom string                          `json:"inPlaceUpdatedFrom,omitempty"`
```
glaksh100, line 187: Minor: Can we add a `Hash` suffix to this?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -1115,6 +1271,19 @@ func (s *FlinkStateMachine) isIncompatibleDeploymentModeChange(application *v1be
 	return application.Spec.DeploymentMode != application.Status.DeploymentMode
 }
 
+// Returns true if the only change between the current status and spec is an _increase_ in parallelism
+func isScaleUp(app *v1beta1.FlinkApplication) bool {
+	if app.Spec.Parallelism > app.Status.JobStatus.Parallelism {
+		appWithOldScale := app.DeepCopy()
+		appWithOldScale.Spec.Parallelism = app.Status.JobStatus.Parallelism
+		hash := flink.HashForApplication(appWithOldScale)
```
glaksh100, line 1279: Is it possible that the deepcopy generates a hash that's != deployHash/failedDeployHash? I seem to have run into this while testing locally. Essentially had an app w. config:
```yaml
spec:
  image: lyft/operator-test-app:b1b3cb8e8f98bd41f44f9c89f8462ce255e0d13f.1
  imagePullSecrets:
    - name: dockerhub
  flinkConfig:
    state.backend.fs.checkpointdir: file:///checkpoints/flink/checkpoints
    state.checkpoints.dir: file:///checkpoints/flink/externalized-checkpoints
    state.savepoints.dir: file:///checkpoints/flink/savepoints
  jobManagerConfig:
    offHeapMemoryFraction: 0.5
    resources:
      requests:
        memory: "100Mi"
        cpu: "0.2"
    replicas: 1
  taskManagerConfig:
    taskSlots: 2
    offHeapMemoryFraction: 0.2
    resources:
      requests:
        memory: "200Mi"
        cpu: "1"
  volumeMounts:
    - mountPath: /checkpoints
      name: checkpoints
  volumes:
    - name: checkpoints
      hostPath:
        path: /tmp/checkpoints
        type: Directory
  flinkVersion: "1.9"
  jarName: "operator-test-app-1.0.0-SNAPSHOT.jar"
  parallelism: 1
  entryClass: "com.lyft.OperatorTestApp"
  scaleMode: "InPlace"
```
Updated parallelism from 1 --> 4 but the generated hash in this method did not match the `DeployHash` (and hence the app never went into a Rescaling state).

Totally possible that it's something w. my local setup but couldn't think of anything that would affect this behaviour (FWIW I re-ran codegen)

mwylde: Did you make sure that only the parallelism changed? You can't use spcli for this currently, as it sets annotations that change the hash.

I think the easiest way to test is to use `kubectl edit` directly.

glaksh100: I was using `kubectl apply`. `kubectl edit` seems to work correctly, though unclear why that would be the case b/c I assumed underneath both commands use PATCH to update resources.

mwylde: Kiubectl apply actually modifies the annotations as well, in order to track which fields it controls.

mwylde: I guess expanding on this a bit, it unfortunate that this feature is not compatible with changes to annotations of the FlinkApplication resource. This happens because we made the decision early on to propagate all annotations on the FlinkApplication to the underlying deployments and pods.

But the pods themselves are immutable, so if we can't do an in-place deploy that involves changes to annotations. If we wanted to support this correctly, we'd have to separate out the annotations on the FlinkApplication resource from the annotations that we set on the deployments and pods.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -720,9 +860,25 @@ func (s *FlinkStateMachine) handleApplicationRunning(ctx context.Context, applic
...
+
+		// if our scale mode is InPlace and this is a suitable update (the only change is an increase in parallelism)
+		// move to Rescaling
+		if application.Spec.ScaleMode == v1beta1.ScaleModeInPlace && isScaleUp(application) {
+			if application.Spec.DeploymentMode == v1beta1.DeploymentModeBlueGreen {
+				s.flinkController.LogEvent(ctx, application, corev1.EventTypeWarning, "NotScalingInPlace",
+					"Not using configured InPlace scaling mode because it is incompatible with BlueGreen"+
+						" deploy mode")
+				s.updateApplicationPhase(application, v1beta1.FlinkApplicationUpdating)
+			} else {
```
glaksh100, line 873: I think this `else` clause should be on one outer-level? Otherwise now, I see the `Application scale has increased. Moving to Rescaling.` even when scaling down.

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -250,6 +256,21 @@ func (s *FlinkStateMachine) handleNewOrUpdating(ctx context.Context, application
 		// Reset teardown hash if set
 		application.Status.TeardownHash = ""
 	}
+
+	// Reset the in-place update hash if set
+	hash := flink.HashForApplication(application)
+	if application.Status.InPlaceUpdatedFrom == hash {
```
glaksh100, line 262: Should we also check for `scaleMode` here and only apply it to the `InPlace` mode? The (possibly contrived) scenario I'm thinking of is when a user moves from 
- P -> P' using  the `InPlace` mode 
- Makes a subsequent update to `NewCluster` mode
- Moves back to P. 

I _think_ we'd still block that according to this condition right?

pkg/controller/flink/job_manager_controller.go
```diff
@@ -92,6 +93,8 @@ func (j *JobManagerController) CreateIfNotExist(ctx context.Context, application
 	newlyCreated := false
 
 	jobManagerDeployment := FetchJobMangerDeploymentCreateObj(application, hash)
+	podSelector := jobManagerDeployment.Spec.Selector.MatchLabels[PodDeploymentSelector]
```
glaksh100, line 96: The pod-deployment-selector on the JM is different from the TM deployment. I'm guessing this doesn't really matter, right? Do we require this label on the JM?

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: The difference in behavior of `kubectl apply` makes sense now. Thanks!

I agree that this is a bit non-ideal. It maybe worth adding a note about the incompatibility to docs/release notes. Also, I realized that with this change, pods can no longer be queried using the `flink-app-hash` label? I don't think this would have been pervasively used, but also maybe useful to document.

mwylde: Added a note about this to the CRD documentation. Longer-term (especially if we're thinking about releasing v1 of the CRD) we may want to look at changing this decision around labels and annotations

`flink-app-hash` is now an annotation on the pods, rather than a label, so it can't be queried in the same way. It's also not going to be consistent across pods in the case of an in-place deploy. I've always considered the app hashes an implementation detail of the operator, and not something that users should be relying on. I don't think we document the current behavior anywhere, so I think this is ok to change.

pkg/controller/flinkapplication/flink_state_machine.go
mwylde: The first condition (on line 867) should only be true if this is a scaleUp, as implemented in `isScaleUp`, which checks that this is an increase in parallelism. If it's not a scale up, we should run the else block on line 878. The inner condition is just about checking for blue/green mode and giving a helpful error message in that case.

pkg/controller/flink/job_manager_controller.go
mwylde: Yeah, it doesn't matter. We just need _something_ to tie the pods to the deployment (and for the jobmanagers, to the services).

pkg/controller/flinkapplication/flink_state_machine.go
mwylde: Yeah, good catch. We should only block if scaleMode is InPlace. I've added that check.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: You're right. My bad.

pkg/controller/flinkapplication/flink_state_machine.go
glaksh100: Why is that the case? Intuitively it seems like if the current scaleMode is NewCluster, `InPlaceUpdatedFrom` should become irrelevant right? And the deployment should be able to proceed as a regular update?

mwylde: I tried to explain the problem in the comment below, but it's a bit complicated. This is basically an edge case when you go from P -> P' -> P, where the P -> P' was in-place. That means that our cluster (e.g., our deployments) are are associated with Hash(P). Now we're trying to go _back_ to P (our original parallelism) which means that we need to create a new set of deployments for Hash(P). But we can't do that, because they already exist (and we're using them). We also can't scale down the existing ones (trying to "recreate" the original setup) because that would disrupt the original job.

The only way to fix the situation is to get the user to change something in the spec that will give us a new hash, allowing the deployment to proceed. Hopefully this is an obscure enough case that users won't hit it, but if they do we need to protect against it.

glaksh100: Okay that makes sense! Thanks for explaining. I wonder if adding `scaleMode` to the application hash would be a way to automatically get us a new hash (at least for the above scenario).

mwylde: Yeah I don't think adding scaleMode into the hash is a good idea, as that would prevent you from enabling in-place and doing an in-place deploy without causing a full deploy to happen.

PR #211, opened 2020-08-04
Remove old flink-app-hash selector from generic service

mwylde:
#209 changed how we associate pods with services; previously we used the `flink-app-hash` label, but in order to decouple hashes from deployments this was changed to the random `pod-deployment-selector`.

For versioned services, this change did not cause any issues, as new versioned services are created for each deploy. The generic (non-versioned) service, however is kept for the lifetime of an application and so needs to be "upgraded" to the new way of selecting pods. #209 included support for upgrading existing services by adding the `pod-deployment-selector` label. However, it did not remove the existing `flink-app-hash` label. This broke generic services for applications that had been deployed on a previous version of the operator.

This PR fixes that, by removing the flink-app-hash label during the deploy process.


anandswaminathan:
Looks good. Want to add a unit test. Very hard to catch if someone removes this.

mwylde:
It's actually fine if this gets removed in the future. It's only needed immediately after upgrading the operator from pre-#209.

PR #217, opened 2020-10-05
Increase job submission timeout and fast fail when it occurs

mwylde:
If the job submission POST request hits our 1 minute timeout (typically because the main method of the job, which needs to run to produce the job graph before job submission can complete) we can end up with two jobs submitted to the cluster.

The sequence of events looks like this:

1. We POST to /run
2. The POST times out
3. We retry the submission, POSTing to /run
4. The JobManager finishes job submission from the first request
5. The JobManager finishes job submission from the second request

Given that, after a timeout, we have no way to know if the job submission is still in progress on the JM, the only safe behavior in this case is to fail the deploy and rollback to the previous version.

This PR adopts that behavior, and also increases the timeout to 5 minutes which should hopefully make this much rarer in practice.

### DIFF
--- a/pkg/controller/flink/client/api.go
+++ b/pkg/controller/flink/client/api.go
@@ -242,7 +242,7 @@ func (c *FlinkJobManagerClient) SubmitJob(ctx context.Context, url string, jarID
 			// up submitted. If we also try submitting again, we may end up with two jobs in the cluster.
 			// The only safe thing to do at this point is give up and allow the user to retry the deploy.
 			return nil, GetNonRetryableError(errors.Errorf(
-				"Job submission timed out after %d seconds, this may be due to the job main method taking too" +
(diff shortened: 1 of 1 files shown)


premsantosh:
👍

glaksh100 (approved):

pkg/controller/flink/client/api.go
```diff
@@ -36,7 +37,7 @@ const httpPost = "POST"
 const httpPatch = "PATCH"
 const retryCount = 3
 const httpGetTimeOut = 5 * time.Second
-const defaultTimeOut = 1 * time.Minute
+const defaultTimeOut = 5 * time.Minute
```
rjbordon, line 40: is there a way to configure timeouts for different operations? I can see that for certain environments (minikube for instance) timeouts are prevalent.

PR #219, opened 2020-10-15
[STRMCMP-1133] Fix for handling nil deployments

premsantosh:
**Issue**
There was a bug when the operator is trying to deploy an app when the previous deployment didn't work. The fetch for deployments resulted in `nil` with in-turn caused the operator to panic because the lack of handling of the `nil` case.

**Fix**
Added a check for `nil` deployments which now returns an error and fails the deployment.


mwylde (approved):
Thanks for finding and fixing!

premsantosh:
🚀

PR #223, opened 2020-11-13
Fix regression in handling of default heap memory config

mwylde:
This fixes a regression introduced in #207 which caused us to use the wrong default for heap memory configs in Flink <1.11.

It also simplifies the memory configuration situation a bit. Now, for Flink <1.11 you may only configure memory with the existing `offHeapMemoryFraction` (which translates into `jobmanager.heap.size`/`taskmanager.heap.size`) and for version >= 1.11 you may only use the new `systemMemoryFraction` (which translates to `jobmanager.memory.process.size`/`taskmanager.memory.process.size`). Using an incompatible configuration for the specified `flinkVersion` will produce an error.

### DIFF
--- a/pkg/controller/flink/config.go
+++ b/pkg/controller/flink/config.go
@@ -108,19 +108,24 @@ func computeMemory(memoryInBytes float64, fraction float64) string {
 	return fmt.Sprintf("%dk", kbs)
 }
 
-func getTaskManagerMemory(app *v1beta1.FlinkApplication, fraction float64) string {
+func getTaskManagerMemory(app *v1beta1.FlinkApplication) string {
 	tmMemory := float64(getRequestedTaskManagerMemory(app))
+	//nolint // fall back to the old config for backwards-compatibility
+	fraction := getFraction(app.Spec.TaskManagerConfig.SystemMemoryFraction, app.Spec.TaskManagerConfig.OffHeapMemoryFraction)
 	return computeMemory(tmMemory, fraction)
 }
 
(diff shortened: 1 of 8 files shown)


pkg/controller/flink/config.go
```diff
@@ -149,11 +152,23 @@ func renderFlinkConfig(app *v1beta1.FlinkApplication) (string, error) {
...
+			return "", fmt.Errorf("systemMemoryFraction config cannot be used with flinkVersion < 1.11', use " +
+				"offHeapMemoryFraction instead")
+		}
+
+		(*config)["jobmanager.heap.size"] = getJobManagerHeapMemory(app)
+		(*config)["taskmanager.heap.size"] = getTaskManagerHeapMemory(app)
 	} else {
-		(*config)["jobmanager.memory.process.size"] = getJobManagerMemory(app)
-		(*config)["taskmanager.memory.process.size"] = getTaskManagerMemory(app)
+		// if heap memory is used for >= 1.11, error out
```
glaksh100, line 163: Not intended to be solved in this PR, but I'm wondering if we can support various flink version --> config compatibility in a different way than `if .. else` blocks. I guess one way could be to couple operator versions to flink versions, but that seems a bit restrictive.

glaksh100 (approved):
Thanks for fixing!

pkg/controller/flink/config.go
mwylde: I don't think we would want to couple operator versions to flink versions, as most deployments will likely be heterogenous (currently at Lyft we have the operator managing applications with Flink 1.9, 1.10, and 1.11).

PR #227, opened 2021-01-06
Updates CRD docs to accurately reflect the behavior of 'savepointpath'.

HunterEl:
Spawned from a question in the community slack channel for this project: [Link to slack conversation](https://flinkk8soperator.slack.com/archives/CMKK848EQ/p1603400167017100).

Specifying `savepointpath` as an update to an already running cluster is treated as a noop, this wasn't clear from the documentation and took us some time to figure this one out. Would be nice if the documentation clearly stated the behavior of this field and how it's expected to behave.

### DIFF
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -103,6 +103,8 @@ Below is the list of fields in the custom resource and their description:
 
   * **savepointPath** `type:string`
     If specified, the application state will be restored from this savepoint.
+    
+    **This will only have an effect on the first deploy of a cluster. Specifying a savepoint path as an update to an already running cluster is treated as a noop.**
 
   * **savepointDisabled** `type:boolean`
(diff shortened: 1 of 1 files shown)


docs/crd.md
```diff
@@ -103,6 +103,8 @@ Below is the list of fields in the custom resource and their description:
 
   * **savepointPath** `type:string`
     If specified, the application state will be restored from this savepoint.
+    
+    **This will only have an effect on the first deploy of a cluster. Specifying a savepoint path as an update to an already running cluster is treated as a noop.**
```
maghamravi, line 107: nit:  What are your thoughts on `s/cluster/application/g` ?  The reason is the existing documentation talks in terms `application` all along and using the same term minimizes confusion.

PR #274, opened 2023-03-13
Use Go Modules

danielmmetz:
And while we're at it, upgrade to Go 1.20. In the process I had to change a lot of the dependencies. Some because the dependencies wouldn't resolve using Go modules, some because they were incompatible with others. The long and the short of it is: 1) the kubernetes dependency ecosystem is very rough and 2) this should be carefully validated during rollout given the number of dependency shifts. 

Some other minor notes include some lint changes, or rewriting a very small bit of code to drop a dependency to have one less to fight with.


anandswaminathan:
@danielmmetz wondering what's happening with integration tests.

anandswaminathan:
PTAL @leoluoInSea

.github/workflows/actions.yml
```diff
@@ -21,9 +21,7 @@ jobs:
       - name: install go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.12
-      - name: install
-        run: make install
+          go-version: "1.20"
```
leoluoInSea, line 24: is the double quote necessary?

leoluoInSea:
> @danielmmetz wondering what's happening with integration tests.

The integ tests are failing. I was trying to debug this and fix the issue, I was able to narrow down the problem is Flink app is stuck at deleting after every test till test time out. However I was unable to figure out why this happened and how to fix it yet.

.github/workflows/actions.yml
danielmmetz: Yes—without it, YAML interprets this as "1.2" and you get a very old version of Go.

sethsaperstein-lyft (approved):

PR #284, opened 2023-03-31
Monitor Job Vertices State on Deploy

leoluoInSea:
# Context
Currently flinkk8soperator only monitor Flink job level overview status, and if the job status is RUNNING, it considers the deployment is succeeded. 

However we encountered scenarios that some vertices are in bad states, the job keeps crash and restart. So the change is also check all vertices status.

[Design doc]( https://docs.google.com/document/d/1m9dUKtIvR8NDZZ9le-CYGEmeuod20vXSL73OiQiP1j8/edit)

# Implementation 
1. ~~Use a feature flag to control the feature ON/OFF. Initially default value is false. After we have enough confidence, we will flip the flag or remove the feature flag.~~ As discussed with Anand, TLDR is this change is relatively small and I will do handful manual testing in staging after merge.
2. I add a logic to fail fast if there is any vertex status is `FAILED`.
3. Add several unit tests
4. I was trying to add a integration test which will add a `,` in `programArgs` and assume it will break args parsing. However, currently test flink app is pinned to a specific old image. I tried to build a new flink image, but new image exceeded memory limit.

# Jira link
https://jira.lyft.net/browse/STRMCMP-1640


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -28,9 +28,10 @@ import (
 )
 
 const (
-	jobFinalizer    = "job.finalizers.flink.k8s.io"
-	statusChanged   = true
-	statusUnchanged = false
+	jobFinalizer                  = "job.finalizers.flink.k8s.io"
+	statusChanged                 = true
+	statusUnchanged               = false
+	jobVertexStateTimeoutInMinute = 3
```
anandswaminathan, line 34: Can this be part of the config so that it can be set through configmap?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -755,33 +756,56 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
-		jobStatus.ProgramArgs = app.Spec.ProgramArgs
-		jobStatus.AllowNonRestoredState = app.Spec.AllowNonRestoredState
-		s.flinkController.UpdateLatestJobStatus(ctx, app, jobStatus)
-		// Update the application status with the running job info
-		app.Status.SavepointPath = ""
-		app.Status.SavepointTriggerID = ""
-		if v1beta1.IsBlueGreenDeploymentMode(app.Status.DeploymentMode) && app.Status.DeployHash != "" {
-			s.updateApplicationPhase(app, v1beta1.FlinkApplicationDualRunning)
+	// wait until all vertices have been scheduled and running
+	jobStartTimeSec := job.StartTime / 1000
```
anandswaminathan, line 760: Is the job start time reliable ? I believe this will get reset when the jm restarts or job fails and restarts

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -755,33 +756,56 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
+	jobStartTime := time.Unix(jobStartTimeSec, jobStartTimeNSec).UTC()
+	now := time.Now().UTC()
+	if now.Before(jobStartTime.Add(jobVertexStateTimeoutInMinute * time.Minute)) {
+		allVerticesRunning := true
+		hasFailure := false
+		failedVertexIndex := -1
+		for index, v := range job.Vertices {
+			if v.Status == client.Failed || v.Status == client.Failing {
+				failedVertexIndex = index
+				hasFailure = true
```
anandswaminathan, line 771: Is this a non recoverable failure?

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: Yes, it could be. But I think even we set it through configmap, we still need a default value in code in case that this key is not set in configMap. If we have the use case that flinkk8soperator user want to override the default value later, we can consider to add it then.

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: Hmm, good point. Let me verify it tmr.

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: I don't have too much exp about this. Could some vertices fail and succeed after retry?

leoluoInSea:
ptal @premsantosh @maghamravi #streaming-compute-dev-prs

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: I have verified that the start time will not reset after job restart.

```{"jobs":[{"jid":"70ec861c2ddf20072a1d2e3f1aff3bd7","name":"nmodes","state":"RESTARTING","start-time":1680713677111,"end-time":-1,"duration":1235989,"last-modification":1680714880571,"tasks":{"total":192,"created":0,"scheduled":0,"deploying":0,"running":0,"finished":0,"canceling":0,"canceled":191,"failed":1,"reconciling":0,"initializing":0}}]}
```

The job is in RESTARTING state, but `start-time` is Wednesday, April 5, 2023 9:54:37.111 AM [GMT-07:00](https://www.epochconverter.com/timezones?q=1680713677111) DST, which is the time I triggered the deployment.

pkg/controller/flinkapplication/flink_state_machine.go
anandswaminathan: Correct. You can add the default value as well. 

https://github.com/lyft/flinkk8soperator/blob/12959ee99d4fd7edc5f801e732a8533a2bd82d53/pkg/controller/config/config.go#L1

pkg/controller/flinkapplication/flink_state_machine.go
anandswaminathan: Yeah. I believe so. Real question is what are you addressing here with hasFailure and deployFailed below.

leoluoInSea:
ptal @premsantosh @maghamravi #streaming-compute-dev-prs

sethsaperstein-lyft:
nit: Can you make the PR title a bit more descriptive

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -754,15 +755,52 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
 	if job == nil {
 		return statusUnchanged, errors.Errorf("Could not find job %s", s.flinkController.GetLatestJobID(ctx, app))
 	}
-
-	// wait until all vertices have been scheduled and started
-	allVerticesStarted := true
-	for _, v := range job.Vertices {
-		allVerticesStarted = allVerticesStarted && (v.StartTime > 0)
+	allVerticesRunning := true
+	if app.Spec.BetaFeaturesEnabled {
```
sethsaperstein-lyft, line 759: Are we documenting what features are in beta?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -754,15 +755,52 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
+		// wait until all vertices have been scheduled and running
+		logger.Info(ctx, "Beta features flag is enabled.")
+		jobStartTimeSec := job.StartTime / 1000
```
sethsaperstein-lyft, line 762: Can we throw the below logic in a separate function to make this more readable and testable?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -754,15 +755,52 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
+					break
+				}
+				allVerticesRunning = allVerticesRunning && (v.StartTime > 0) && v.Status == client.Running
+			}
+			// fail fast
+			if hasFailure {
+				s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobRunningFailed",
+					fmt.Sprintf(
+						"Vertex %d with name [%s] state is Failed", failedVertexIndex, job.Vertices[failedVertexIndex].Name))
+				return s.deployFailed(app)
```
anandswaminathan, line 782: This is a new behavior. Is it intentional ? More than monitoring, we are updating the state of the deployment here

pkg/controller/flinkapplication/flink_state_machine.go
anandswaminathan: This is not a standard pattern.

Primarily because, this flag getting updated will cause applications to restart.

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: This change is not just monitoring but also updating state of deployment when conditions are met. 
This specific fail fast part is new and it's based on general philosophy that if fails, fail fast. 

I double checked Flink doc about the vertex state transition graph, `FAILING` could retry while `FAILED` is the final bad state. 

So I will update the logic to if any vertex state is `FAILED`, fail deployment fast.

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: @sethsaperstein-lyft I will add more to crd.md file for documenting

leoluoInSea: @anandswaminathan sorry, I didn't follow that why if the flag is updated, the application to restart. Can you elaborate that? My understanding is that this field is set in applicaiton jsonnet file and the value is passed in when new deployment/update is made. But this field can't be changed in runtime.

leoluoInSea:
ptal @sethsaperstein-lyft @anandswaminathan

pkg/controller/flinkapplication/flink_state_machine.go
sethsaperstein-lyft: I should have clarified that I meant all the new logic of the running vertex state check, not just the start time calculation

pkg/controller/flinkapplication/flink_state_machine.go
anandswaminathan: @leoluoInSea 

So here is what you are doing - you are looking for a way to have some applications be exposed to this feature. 

This is not something that is commonly done in CRD's and operator, as this is not something you want to have for the long run. Say you remove the flag some day later, all the apps that have this set need to revert, and we are looking at updating 100s of apps (depending on scale). 

If you are worried about the reliability of the change, may be pass a version as label (if we already don't do so), and have logic on the label. That way, if the code gets removed, or changed, the apps will not need to be updated.

leoluoInSea: Just for record, Anand and I had a discussion and our conclusion is that this is not breaking changes, and I will do several testing in staging after merging, definitely with an example of failure scenario `nmodes`. 

For future benefit if we need a feature flag, as Anand pointed out above, the recommended way is to check label instead of adding a new field in CRD. From app jsonnet file, just add the label if we need to turn on the feature.

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: After removing the feature flag if condition, I think it's readable now.

leoluoInSea:
ptal @anandswaminathan

pkg/controller/flinkapplication/flink_state_machine.go
sethsaperstein-lyft: func is >100 lines. I think it's somewhat difficult to read. Nit, but for maintainability

leoluoInSea:
ptal @anandswaminathan

anandswaminathan (approved):
Please get a +1 from @sethsaperstein-lyft

sethsaperstein-lyft (approved):

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -756,14 +756,58 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
+	logger.Info(ctx, "Job vertex timeout config is ", cfg.FlinkJobVertexTimeout)
+	flinkJobVertexTimeout := cfg.FlinkJobVertexTimeout
+	if now.Before(jobStartTime.Add(flinkJobVertexTimeout.Duration)) {
+		allVerticesRunning, hasFailure, failedVertexIndex := monitorAllVerticesState(job)
+
+		// fail fast
+		if hasFailure {
+			s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobRunningFailed",
+				fmt.Sprintf(
+					"Vertex %d with name [%s] state is Failed", failedVertexIndex, job.Vertices[failedVertexIndex].Name))
```
premsantosh, line 771: Curious, will this log message make sense to users?

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -756,14 +756,58 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
+	// wait until all vertices have been scheduled and running
+	hasFailure := false
+	failedVertexIndex := -1
+	for index, v := range job.Vertices {
+		if v.Status == client.Failed {
+			failedVertexIndex = index
+			hasFailure = true
+			break
+		}
+		allVerticesRunning = allVerticesRunning && (v.StartTime > 0) && v.Status == client.Running
```
premsantosh, line 795: If the first vertex is failing then won't this return `allVerticesRunning` as `true` when it should be `false`?

leoluoInSea: If the vertex is failing, `hasFailure` is set to true, then line 768 if block will have early return to fail the deployment.

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: An example of the message is like below. I think it gives hints to point out which vertex may has problem

```
Vertex 2 with name [[9]{Process events with stateful dofn, nmodes_filter_15m_data, filter_active_sessions, nmodes_timestamp, __aggregate_event_processing_results_timestamp, nmodes_window_15m, __aggregate_event_processing_results_filter_up5m_data, __aggregate_event_processing_results_window_1m, __aggregate_event_processing_results_window_5m} -> ToKeyedWorkItem] state is Failed
```

premsantosh: Thanks!

pkg/controller/flinkapplication/flink_state_machine.go
leoluoInSea: Agree. Initially this part is in the main method, so I didn't set value for `allVerticesRunning` in fail scenario. I will provide the fix later

pkg/controller/flinkapplication/flink_state_machine.go
sethsaperstein-lyft: @anandswaminathan in retrospect this behavior did not follow the original design as we would like to move to rollingback. New PR: https://github.com/lyft/flinkk8soperator/pull/291

anandswaminathan:
Checking.

PR #286, opened 2023-04-13
fix a potential bug

leoluoInSea:
Fix a potential bug. If any vertex is failed, also need to set `allVerticesRunning` parameter to false explicitly. 

Currently this will not cause any issue because caller ignores `allVerticesRunning` if `hasFailure` is true. 

But future caller may make wrong assumption.

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -790,6 +790,7 @@ func monitorAllVerticesState(job *client.FlinkJobOverview) (bool, bool, int) {
 		if v.Status == client.Failed {
 			failedVertexIndex = index
 			hasFailure = true
+			allVerticesRunning = false
 			break
 		}
 		allVerticesRunning = allVerticesRunning && (v.StartTime > 0) && v.Status == client.Running


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -790,6 +790,7 @@ func monitorAllVerticesState(job *client.FlinkJobOverview) (bool, bool, int) {
 		if v.Status == client.Failed {
 			failedVertexIndex = index
 			hasFailure = true
+			allVerticesRunning = false
```
premsantosh, line 793: If I understand correctly you don't need line 796

leoluoInSea: Do you mean `allVerticesRunning = allVerticesRunning && (v.StartTime > 0) && v.Status == client.Running`? We still need this right? Vertex state could be other than RUNNING, like `FAILING` or `RESTARTING`. In these cases, vertex is restarting itself, operator should keep monitoring and wait for next check.

There are three cases
1. All vertices are RUNNING, this is happy path, operator mark the deployment as successful
2. Any vertex is FAILED, which is final failed state and vertex will not retry or recover, so operator mark the deployment as failed.
3. Other than 1 and 2, operator will not change the deployment status, and wait for next scan.

leoluoInSea:
ptal @premsantosh

pkg/controller/flinkapplication/flink_state_machine.go
premsantosh: You're right!

premsantosh (approved):

PR #291, opened 2023-05-15
[STRMHELP-315] Rollback on Failed Job Monitoring 🐛 

sethsaperstein-lyft:
## overview
In the [job monitoring PR](https://github.com/lyft/flinkk8soperator/pull/284) we introduced a bug such that when the job monitoring fails due to timeout or a failed vertex, the state DeployFailed is reached instead of attempting to rollback. This simplifies the logic of job monitoring as well as resulting in the job attempting to roll back

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -2,6 +2,7 @@ package flinkapplication
 
 import (
 	"context"
+	config2 "github.com/lyft/flytestdlib/config"
 	"math"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -767,47 +768,38 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
-		// fail fast
-		if hasFailure {
-			s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobRunningFailed",
-				fmt.Sprintf(
-					"Vertex %d with name [%s] state is Failed", failedVertexIndex, job.Vertices[failedVertexIndex].Name))
-			return s.deployFailed(app)
-		}
-		return updateJobAndReturn(ctx, job, s, allVerticesRunning, app, hash)
+	logger.Info(ctx, "Monitoring job vertices with timeout ", flinkJobVertexTimeout)
+	jobStarted, err := monitorJobStart(job, flinkJobVertexTimeout)
```
maghamravi, line 774: nit: would be good to call the method name `monitorJobSubmission` and `jobStarted` to `status`

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -767,47 +768,38 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
...
+	}
+	return statusUnchanged, err
 }
 
-func monitorAllVerticesState(job *client.FlinkJobOverview) (bool, bool, int) {
-	allVerticesRunning := true
-	// wait until all vertices have been scheduled and running
-	hasFailure := false
-	failedVertexIndex := -1
+func monitorJobStart(job *client.FlinkJobOverview, timeout config2.Duration) (bool, error) {
```
maghamravi, line 786: nice to see this method being succinct now !

maghamravi (dismissed):

pkg/controller/flinkapplication/flink_state_machine.go
maghamravi: There are **only** two hard things in Computer Science: cache invalidation and naming things :)  
My rationale to rename `jobStarted` to `status` was primarily due my other suggestion for renaming the method to `monitorJobSubmission`.  Given the method was returning a boolean, a status felt more natural.  Okay to keep it as `jobStarted`.

maghamravi (approved):

PR #298, opened 2023-12-18
Change app status to rollback on job start timeout

avneet15:
Once a new Flink job is submitted, it’s possible that the job vertices are not in the running state by the configured timeout period of 3 minutes.

There exists a bug in the flinkk8scontroller where the app will get stuck in this phase, instead of being rolled back and marked as DeployFailed.

Slack [thread](https://lyft.slack.com/archives/CPN3NG47M/p1701461663432759?thread_ts=1701458116.814489&cid=CPN3NG47M)

### DIFF
--- a/pkg/controller/flinkapplication/flink_state_machine.go
+++ b/pkg/controller/flinkapplication/flink_state_machine.go
@@ -784,6 +784,7 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
 	if err != nil {
 		logger.Info(ctx, "Job monitoring failed with error: %v", err)
 		s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobMonitoringFailed", err.Error())
+		s.flinkController.UpdateLatestJobID(ctx, app, "")
 		s.updateApplicationPhase(app, v1beta1.FlinkApplicationRollingBackJob)
 		return statusChanged, err
 	}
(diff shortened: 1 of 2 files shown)


avneet15:
👀 @sethsaperstein-lyft

pkg/controller/flinkapplication/flink_state_machine.go
```diff
@@ -784,7 +784,8 @@ func (s *FlinkStateMachine) handleSubmittingJob(ctx context.Context, app *v1beta
 	if err != nil {
 		logger.Info(ctx, "Job monitoring failed with error: %v", err)
 		s.flinkController.LogEvent(ctx, app, corev1.EventTypeWarning, "JobMonitoringFailed", err.Error())
-		return statusUnchanged, err
+		s.updateApplicationPhase(app, v1beta1.FlinkApplicationRollingBackJob)
```
sethsaperstein-lyft, line 787: Just read through the code and it seems we should set the job ID to "" similar to above
https://github.com/lyft/flinkk8soperator/pull/298/files#diff-4322066c6bf591e2dc686197f8d021859e75877db2e635e1a1c3b83abdd660f8L717

From the code it seems that after setting rolling back that it goes to a rollback handler which only submits the job if the job id in status is "": https://sourcegraph.lyft.net/github.com/lyft/flinkk8soperator/-/blob/pkg/controller/flinkapplication/flink_state_machine.go?L884

I'm not a fan of this connected logic but again we're looking to deprecate this operator in Q1 so I digress

avneet15: Nice catch, would definitely be preferable to re-submit job if possible. Will set this to be empty.

avneet15:
👀 @sethsaperstein-lyft @lyft/async-platform

sethsaperstein-lyft (approved):

avneet15:
🚀 ✅

avneet15:
🚀 ✅

PR #299, opened 2024-01-09
Headless Service

sethsaperstein-lyft:
## overview
Change services to headless service as ClusterIP is not required. Apache OSS operator also uses headless services.


premsantosh (approved):

maheepm-lyft (approved):
lgtm

